PR #59, opened 2019-07-25
[#20] Switch to checking generated crates

### LINKED ISSUE #20: Switch to checking generated crates
We're only interested in whether the code we've generated actually compiles. We don't need to build the artifacts. Let's change `.travis.yml` and `Makefile` to `cargo check` `test_k8s` and `test_pet` crates (instead of `cargo build`).

rabotyaga:
closes #20


wafflespeanut:
Thank you! :smile:

PR #65, opened 2019-07-31
[#29] Support request timeouts in generated CLI

### LINKED ISSUE #29: Support request timeouts in generated CLI
`reqwest` supports setting [timeouts](https://docs.rs/reqwest/0.9.18/reqwest/async/struct.ClientBuilder.html#method.timeout) for HTTP requests. We could accept that as an optional argument in CLI and set it (if needed) [before building](https://github.com/wafflespeanut/paperclip/blob/25f59b63fac9b7aa3a66521960a65a8311984442/src/v2/codegen/state.rs#L553) the client.

rabotyaga:
closes #29

### DIFF
--- a/src/v2/codegen/state.rs
+++ b/src/v2/codegen/state.rs
@@ -405,6 +405,11 @@ args:
         short: v
         long: verbose
         help: Enable verbose mode.
+    - timeout:
+        short: t
+        long: timeout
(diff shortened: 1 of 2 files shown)


wafflespeanut:
Thanks for your PR! :)

Issue #108, opened 2019-10-01
Add missing fields to items object


wafflespeanut:
OpenAPI [items object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#items-object) has a number of fields which we currently don't support. The first step of #101 is to add the missing fields to [`Items` struct](https://github.com/wafflespeanut/paperclip/blob/315a8050f04d13f02aff7932b21e65b0e7e921d6/core/src/v2/models.rs#L324). The missing fields are all `Option`al and are mostly simple types (exception being `default` field, which should be `serde_json::Value`).

wafflespeanut:
This is being worked on by @Sobert

Sobert:
On it (will finish it tomorrow morning and try to properly do another one now that I have more insight).

PR #110, opened 2019-10-01
#108 added missing fields to items object

Sobert:
Hopefully I understood correctly what was needed.

### DIFF
--- a/core/src/v2/models.rs
+++ b/core/src/v2/models.rs
@@ -332,6 +332,28 @@ pub struct Items {
     pub collection_format: Option<CollectionFormat>,
     #[serde(default, rename = "enum", skip_serializing_if = "Vec::is_empty")]
     pub enum_: Vec<serde_json::Value>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub maximum: Option<f32>,
+    #[serde(rename = "exclusiveMaximum", skip_serializing_if = "Option::is_none")]
+    pub exclusive_maximum: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub minimum: Option<f32>,
+    #[serde(rename = "exclusiveMinimum", skip_serializing_if = "Option::is_none")]
+    pub exclusive_minimum: Option<bool>,
+    #[serde(rename = "maxLength", skip_serializing_if = "Option::is_none")]
+    pub max_length: Option<u32>,
+    #[serde(rename = "minLength", skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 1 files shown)


wafflespeanut:
Please add a comment in the issue saying you're willing to work on it, so that I can assign it to you and we can avoid duplicate efforts for someone else.

core/src/v2/models.rs
```diff
@@ -332,6 +332,28 @@ pub struct Items {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub maximum: Option<f32>,
+    #[serde(rename = "exclusiveMaximum", skip_serializing_if = "Option::is_none")]
+    pub exclusive_maximum: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub minimum: Option<f32>,
+    #[serde(rename = "exclusiveMinimum", skip_serializing_if = "Option::is_none")]
+    pub exclusive_minimum: Option<bool>,
+    #[serde(rename = "maxLength", skip_serializing_if = "Option::is_none")]
+    pub max_length: Option<i32>,
```
wafflespeanut: Even though the spec says this is an integer, it also says that the integer must be greater than or equal to zero, in which case, we should go for the unsigned version (`u32`).

core/src/v2/models.rs
```diff
@@ -332,6 +332,28 @@ pub struct Items {
...
+    #[serde(rename = "minLength", skip_serializing_if = "Option::is_none")]
+    pub min_length: Option<i32>,
```
wafflespeanut: Same as `max_length`.

core/src/v2/models.rs
```diff
@@ -332,6 +332,28 @@ pub struct Items {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub pattern: Option<String>,
+    #[serde(rename = "maxItems", skip_serializing_if = "Option::is_none")]
+    pub max_items: Option<i32>,
```
wafflespeanut: Same as the above.

core/src/v2/models.rs
```diff
@@ -332,6 +332,28 @@ pub struct Items {
...
+    #[serde(rename = "minItems", skip_serializing_if = "Option::is_none")]
+    pub min_items: Option<i32>,
```
wafflespeanut: Same as the above.

Sobert:
> Please squash all your commits once you've made the changes. :)

It should be all good now :)

wafflespeanut:
Thanks!

PR #111, opened 2019-10-01
Parameter object

### LINKED ISSUE #109: Add missing fields to parameter object
OpenAPI [parameter object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject) has some fields which we currently don't support. Let's add the missing fields to [`Parameter` struct](https://github.com/wafflespeanut/paperclip/blob/315a8050f04d13f02aff7932b21e65b0e7e921d6/core/src/v2/models.rs#L295). The missing fields are all `Option`al and are mostly simple types (exception being `default` field, which should be `serde_json::Value`).

iambenzo:
Fixes #109

Please review and let me know if i need to update any of the data types - After reading into the data types required, I used `u32` for the `integer` values and `f32` for the `number` values.

### DIFF
--- a/core/src/v2/actix.rs
+++ b/core/src/v2/actix.rs
@@ -120,6 +120,18 @@ where
             items: None,
             enum_: Default::default(),
             collection_format: None,
+            default: None,
+            maximum: None,
+            exclusive_maximum: None,
+            minimum: None,
+            exclusive_minimum: None,
+            max_length: None,
+            min_length: None,
+            pattern: None,
+            max_items: None,
+            min_items: None,
+            unique_items: None,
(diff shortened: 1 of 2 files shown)


wafflespeanut:
Thanks for working on this! The existing changes look okay, but it looks like the code doesn't compile anymore. Can you attempt to fix it? Also, please squash the commits after making the relevant changes.

iambenzo:
Oh dear! I will work on this, indeed.

iambenzo:
Squashed :)

wafflespeanut:
Thank you! :smile:

PR #112, opened 2019-10-01
Add support for date-time format (fixes #107)

### LINKED ISSUE #107: Support date-time format behind feature gate
The crate `chrono` offers a type for datetime support and since it's [already a known format](https://github.com/wafflespeanut/paperclip/blob/315a8050f04d13f02aff7932b21e65b0e7e921d6/core/src/v2/models.rs#L74), we can implement it behind a feature gate.

 - Add `chrono = 0` (with `optional = true`) in [`Cargo.toml` of `core` crate](https://github.com/wafflespeanut/paperclip/blob/315a8050f04d13f02aff7932b21e65b0e7e921d6/core/Cargo.toml).
- Add a new feature named `chrono` with `chrono` in the array (for reference, see `heck` dependency and `codegen` feature which depends on `heck` in the manifest).
- In `schema` module, we already [have a number of implementations](https://github.com/wafflespeanut/paperclip/blob/315a8050f04d13f02aff7932b21e65b0e7e921d6/core/src/v2/schema.rs#L130-L143) using `impl_type_simple!` macro. Add an implementation for `chrono::NaiveDateTime` with `DataType::String` and `DataTypeFormat::DateTime`.
- Add a feature gate before that macro impl. It'll look something like [this](https://github.com/wafflespeanut/paperclip/blob/6e67d1fa107074e0ad6a2e56e38af7bc898a55ca/core/src/v2/mod.rs#L12), but with `chrono` instead of `actix`.

Feel free to ask any questions!

wyze:


wafflespeanut:
Thank you for working on this! :smile:

wyze:
Thank you for the helpful and welcoming experience!

PR #113, opened 2019-10-01
[#106] Support uuid format with feature flag

### LINKED ISSUE #106: Support UUID format behind feature gate
The crate `uuid` offers a type for UUID support and since it's [already a known format](https://github.com/wafflespeanut/paperclip/blob/315a8050f04d13f02aff7932b21e65b0e7e921d6/core/src/v2/models.rs#L76), we can implement it behind a feature gate.

 - Add `uuid = 0` (with `optional = true`) in [`Cargo.toml` of `core` crate](https://github.com/wafflespeanut/paperclip/blob/315a8050f04d13f02aff7932b21e65b0e7e921d6/core/Cargo.toml).
- Add a new feature named `uid` with `uuid` in the array (for reference, see `heck` dependency and `codegen` feature which depends on `heck` in the manifest).
- In `schema` module, we already [have a number of implementations](https://github.com/wafflespeanut/paperclip/blob/315a8050f04d13f02aff7932b21e65b0e7e921d6/core/src/v2/schema.rs#L130-L143) using `impl_type_simple!` macro. Add an implementation for `uuid::Uuid` with `DataType::String` and `DataTypeFormat::Uuid`.
- Add a feature gate before that macro impl. It'll look something like [this](https://github.com/wafflespeanut/paperclip/blob/6e67d1fa107074e0ad6a2e56e38af7bc898a55ca/core/src/v2/mod.rs#L12), but with `uuid` instead of `actix`.
- We need to add tests for this too. Let's add `uuid` to the list of `[dev-dependencies]` in the root manifest. Next stop would be to add a new field (say, `uid`) with `uuid::Uuid` type to `Pet` struct in `tests/test_app.rs`. This will fail existing tests because the JSON will be different. Once we update the expected JSON outputs, the tests will pass again.

Feel free to ask any questions!

hindenbug:
This PR Resolves #106 
- Adds `uuid = 0` (with `optional = true`) in Cargo.toml of core crate.
- Adds `uuid` with `serde` enabled  in Cargo.toml of root crate.
- Add a new feature named `uid` = `[uuid]` to Cargo.toml core crate & `uid = ["paperclip-core/uid"]` in root Cargo.toml
- Add an implementation for `uuid::Uuid` format
- Updates tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ actix-service = "0.4.1"
 actix-web = "1.0.4"
 chrono = { version = "0.4", features = ["serde"] }
 futures = "0.1"
+uuid = { version = "0.7.4", features = ["serde"] }
 
 [features]
 actix = ["paperclip-macros/actix", "paperclip-actix"]
@@ -56,6 +57,7 @@ v2 = ["paperclip-macros/v2", "paperclip-core/v2"]
 codegen = ["heck", "lazy_static", "regex", "tinytemplate", "paperclip-core/codegen"]
 cli = ["default", "env_logger", "structopt", "git2", "reqwest"]
 # codegen-fmt = ["codegen", "rustfmt-nightly"]
+uid = ["paperclip-core/uid"]
 
 [workspace]
 members = [
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ build:
 	cargo build --features default
 	cargo build --features actix
 	cargo build --features cli
+	cargo build --features uid
 	cargo build --all --all-features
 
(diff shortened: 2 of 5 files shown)


wafflespeanut:
This looks great! Thanks for the PR! There's a merge conflict because of #112. Could you rebase on top of it?

hindenbug:
@wafflespeanut  Rebased :)

hindenbug:
Oops, my bad. I think something broke in the tests (will fix it in a while)

wafflespeanut:
Nice work! Thanks! :smile:

PR #114, opened 2019-10-02
fixes #105

### LINKED ISSUE #105: Move collection format validation block to a separate function
There's a [validation block](https://github.com/wafflespeanut/paperclip/blob/315a8050f04d13f02aff7932b21e65b0e7e921d6/src/v2/codegen/emitter.rs#L759-L797) in `RequirementsCollector::collect_parameters` which is quite big. Let's move this block to another method, say `validate_collection_format` below.

ansrivas:
This pull request fixes #105. 
Although this does exactly what was explained in the issue, I am a bit skeptical about passing mutable `Vec<CollectionFormat>`.

In case you'd want to suggest any changes in the approach, please let me know.

### DIFF
--- a/src/v2/codegen/emitter.rs
+++ b/src/v2/codegen/emitter.rs
@@ -710,6 +710,57 @@ where
         Ok(())
     }
 
+    fn validate_collection_format(
+        &mut self,
+        p: &models::Parameter<SchemaRepr<E::Definition>>,
+        it_fmts: &mut Vec<CollectionFormat>,
+    ) {
+        if p.data_type != Some(DataType::Array) {
+            return;
+        }
+        // If it's an array, then validate collection formats and default if needed.
+        let default_fmt = CollectionFormat::default();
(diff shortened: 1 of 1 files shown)


src/v2/codegen/emitter.rs
```diff
@@ -710,6 +710,49 @@ where
         Ok(())
     }
 
+    fn validate_collection_format(&mut self, p:&models::Parameter<SchemaRepr<E::Definition>>,it_fmts:&mut Vec<CollectionFormat>) {
+        // If it's an array, then validate collection formats and default if needed.
+        if p.data_type == Some(DataType::Array) {
```
wafflespeanut: It'd be relatively easier to read if this condition looks like:
``` rust
if p.data_type != Some(DataTy

```

wafflespeanut:
Thanks for working on this! Two things before we can merge this:
- The commit message should be moved to the first line (`fixes #105` could be in the second).
- Could you also run rustfmt and clippy in your code? The CI build is failing.

wafflespeanut:

PR #122, opened 2019-10-03
fixes #73

### LINKED ISSUE #73: Actix: can't use serde_json::Value with api_v2_schema
I have a few responses and requests that need to take any valid JSON as part of a parameter. Up till now I have accomplished this using serde_json::Value. I'm not certain what the best way to implement raw_schema for this would be but obviously it can't be done directly. 

Other thoughts would be is there a way I could annotate an attribute as being Any?

devameen:


tests/test_app.rs
```diff
@@ -288,7 +288,7 @@ fn test_params() {
                     "BadgeBody": {
                       "properties": {
                         "data": {
-                          "type": "string"
+                          "type": {}
                         }
```
wafflespeanut: This should be `"data": {}`

wafflespeanut:
Okay, I've missed something.

Let's do the following:
 - Revert the change in `BadgeBody`
 - Change `BadgeBody` to `BadgeForm` and update its occurrences.
 - Add a new struct `BadgeBody` that looks like this:

```rust
#[api_v2_schema]
#[derive(Deserialize)]
struct BadgeBody {
    json: Option<serde_json::Value>,
    yaml: Option<serde_yaml::Value>,
}
```

Next stop would be to update the tests accordingly.

devameen:
Cool. will make those changes

tests/test_app.rs
```diff
@@ -242,13 +249,13 @@ fn test_params() {
...
         _q: web::Query<BadgeParams>,
-        _f: web::Form<BadgeBody>,
+        _f: web::Form<BadgeForm>,
     ) -> String {
         unimplemented!();
     }
 
     #[api_v2_operation]
-    fn post_badge_2(_p: web::Path<(String, String)>, _b: web::Json<BadgeBody>) -> String {
+    fn post_badge_2(_p: web::Path<(String, String)>, _b: web::Json<BadgeForm>) -> String {
```
wafflespeanut: Can you change this to `BadgeBody`, so that we can actually test what we're trying to do?

devameen: Cool! sure.

wafflespeanut:

Issue #123, opened 2019-10-03
Url in example throws 404


M4rk9696:
The URL https://raw.githubusercontent.com/wafflespeanut/paperclip/master/openapi/tests/pet-v2.yaml in the CLI example throws 404

URL to the Example: https://paperclip.waffles.space/cli.html

wafflespeanut:
Oh yes, the new URL is actually https://raw.githubusercontent.com/wafflespeanut/paperclip/master/tests/pet-v2.yaml and the change lives in `book/cli.md`

Thanks for finding this!

M4rk9696:
I'll raise a PR

wafflespeanut:
I've assigned it to you!

PR #124, opened 2019-10-04
Update link in CLI example

M4rk9696:
Update the link in the CLI example with a valid link for #123

cc/ @wafflespeanut


wafflespeanut:
Thank you very much! :smile:

PR #127, opened 2019-10-04
[#76] Support newtype

### LINKED ISSUE #76: Actix-web: Support 1-tuple structs
At the moment, it doesn't seem like "newtypes" are supported, e.g.

```rust
#[api_v2_schema)]
#[derive(Serialize, Deserialize)]
pub struct TodoId(pub u64);
```

It would be neat if we could support this, perhaps by making it show as the underlying type in the Swagger spec.

hindenbug:
Resolves #76 

- Add support for `newtype` or `1 tuple structs`
- Extract getting the field type logic to a new `fn get_field_type`

### DIFF
--- a/macros/src/actix/mod.rs
+++ b/macros/src/actix/mod.rs
@@ -6,8 +6,8 @@ use proc_macro::TokenStream;
 use quote::quote;
 use syn::spanned::Spanned;
 use syn::{
-    Data, DataEnum, Fields, FieldsNamed, ItemFn, Meta, NestedMeta, PathArguments, ReturnType,
-    Token, TraitBound, Type,
+    Data, DataEnum, Field, Fields, FieldsNamed, FieldsUnnamed, ItemFn, Meta, NestedMeta,
+    PathArguments, ReturnType, Token, TraitBound, Type,
 };
 
 const SCHEMA_MACRO: &str = "api_v2_schema";
@@ -124,12 +124,16 @@ pub fn emit_v2_definition(attrs: TokenStream, input: TokenStream) -> TokenStream
         Data::Struct(ref s) => match &s.fields {
             Fields::Named(ref f) => handle_field_struct(f, &mut props_gen),
             Fields::Unnamed(ref f) => {
-                let s = f.span().unwrap();
-                s.warning(
-                    "tuple structs do not have named fields and hence will have empty schema.",
-                )
-                .emit();
-                s.help(&*EMPTY_SCHEMA_HELP).emit();
+                if f.unnamed.len() == 1 {
+                    handle_unnamed_field_struct(f, &mut props_gen)
+                } else {
+                    let s = f.span().unwrap();
+                    s.warning(
+                        "tuple structs do not have named fields and hence will have empty schema.",
+                    )
+                    .emit();
+                    s.help(&*EMPTY_SCHEMA_HELP).emit();
+                }
             }
             Fields::Unit => {
                 let s = s.struct_token.span().unwrap();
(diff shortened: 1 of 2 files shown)


macros/src/actix/mod.rs
```diff
@@ -124,12 +124,16 @@ pub fn emit_v2_definition(attrs: TokenStream, input: TokenStream) -> TokenStream
         Data::Struct(ref s) => match &s.fields {
             Fields::Named(ref f) => handle_field_struct(f, &mut props_gen),
             Fields::Unnamed(ref f) => {
-                let s = f.span().unwrap();
-                s.warning(
-                    "tuple structs do not have named fields and hence will have empty schema.",
-                )
-                .emit();
-                s.help(&*EMPTY_SCHEMA_HELP).emit();
+                if f.unnamed.len() == 1 {
```
hindenbug: I assumed we check for `newtype` / 1 tuple structs ?

macros/src/actix/mod.rs
```diff
@@ -169,6 +173,36 @@ pub fn emit_v2_definition(attrs: TokenStream, input: TokenStream) -> TokenStream
...
+                .expect("expected type for struct field");
+
+            if p.path.segments.len() == 1 && &ty.ident == "Option" {
+                is_required = Some(false);
+            }
+
+            address_type_for_fn_call(&field.ty)
+        }
+        Type::Reference(_) => address_type_for_fn_call(&field.ty),
+        _ => address_type_for_fn_call(&field.ty),
```
hindenbug: not fully sure about this 🤔

macros/src/actix/mod.rs
```diff
@@ -169,6 +173,36 @@ pub fn emit_v2_definition(attrs: TokenStream, input: TokenStream) -> TokenStream
...
+    };
+}
+
+/// Generates code for a tuple struct with fields.
+fn handle_unnamed_field_struct(fields: &FieldsUnnamed, props_gen: &mut proc_macro2::TokenStream) {
+    let field = fields.unnamed.iter().nth(0).unwrap();
+    let ty_ref = get_field_type(field, None);
+
+    props_gen.extend(quote!({
```
hindenbug: Getting a, not sure why maybe coz of above 🤔 
```
error[E0277]: the trait bound `(): quote::to_tokens::ToTokens` is not satisfied
   --> macros/src/actix/mod.rs:201:22
    |
201 |       props_gen.extend(quote!({
    |  ______________________^
202 | |         schema = #ty_ref::raw_schema();
203 | |     }));
    | |______^ the trait `quote::to_tokens::ToTokens` is not implemented for `()`
    |
    = note: required by `quote::to_tokens::ToTokens::to_tokens`
    = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

macros/src/actix/mod.rs
wafflespeanut: That's correct.

macros/src/actix/mod.rs
```diff
@@ -169,6 +173,36 @@ pub fn emit_v2_definition(attrs: TokenStream, input: TokenStream) -> TokenStream
     gen.into()
 }
 
+fn get_field_type(field: &Field, mut is_required: Option<bool>) {
```
wafflespeanut: This function definition is causing the error. By default, a block (which doesn't evaluate to anything) expresses an empty tuple `()`, which is what we see in the error. Since we're proxying `address_type_for_fn_call`, we should return whatever that function returns (i.e., `proc_macro2::TokenStream`). However, there are cases where we don't know the appropriate type, so we should return an `Option<proc_macro2::TokenStream>`.

One other thing I'd like to address is the `is_required` variable. It doesn't actually have to be an optional, because it's either required or it's not (just a `bool` should work). And since it's being used outside this context, we should either take a mutable reference and mutate it or initialize the variable within this block and return a `bool` along with `Option<..>` as a tuple. Otherwise, `bool` (being a `Copy` type) gets copied whenever we pass it by value.

macros/src/actix/mod.rs
wafflespeanut: This should be `None`, since we don't care other kinds of types just yet. It looks like you've removed the warning we used to emit.

macros/src/actix/mod.rs
```diff
@@ -179,31 +213,7 @@ fn handle_field_struct(fields: &FieldsNamed, props_gen: &mut proc_macro2::TokenS
...
-                address_type_for_fn_call(&field.ty)
-            }
-            Type::Reference(_) => address_type_for_fn_call(&field.ty),
-            _ => {
-                field
-                    .ty
-                    .span()
-                    .unwrap()
-                    .warning("unsupported field type will be ignored.")
-                    .emit();
```
wafflespeanut: This warning should still be emitted, so that users can see that it's something we don't support now.

macros/src/actix/mod.rs
wafflespeanut: We do this only when we get a type. If we don't get a type out of that function, we'll skip extending `props_gen` entirely.

macros/src/actix/mod.rs
hindenbug: Thanks for this 😄 , I completely forgot about the return type `proc_macro2::TokenStream

wafflespeanut:
Looks good. Now, let's test this! We have an [`Image` struct](https://github.com/wafflespeanut/paperclip/blob/eef0a06649da373900df8fcdbc205a27bc6cb96e/tests/test_app.rs#L422) in `tests/test_app.rs`. Let's add an unit struct before it, which looks like this:

```rust
#[api_v2_schema]
#[derive(Deserialize, Serialize)]
struct ImageId(u64);
```

... and a new field called `id`, with this unit struct. Previously, this emits a warning. Now, it shouldn't. You also have to change the expected JSON (below) to get the test to pass. You can see the diff by running the test and comparing the JSON outputs.

wafflespeanut:

hindenbug:
Thanks a lot  for being patient with this :) @wafflespeanut

PR #131, opened 2019-10-10
[#35] Use deprecated field for operations

### LINKED ISSUE #35: Use deprecation field in operation
The `deprecation` field in Operation items specify whether that operation is deprecated. We should mark the corresponding functions (with `deprecated` attribute) in the generated code and issue that as a message during compilation.

hindenbug:
This PR Resolves #35 
-  Adds `deprecated` field to `Operation`, `OpRequirement`
-  Adds generation of `[#deprecated]` attribute in builder, if 'deprecation` value present

### DIFF
--- a/src/v2/codegen/object.rs
+++ b/src/v2/codegen/object.rs
@@ -56,6 +56,8 @@ pub struct OpRequirement {
     pub id: Option<String>,
     /// Description of this operation (if any), to be used for docs.
     pub description: Option<String>,
+    /// Whether the operation is deprecated or not.
+    pub deprecated: bool,
     /// Parameters required for this operation.
     pub params: Vec<Parameter>,
     /// Whether the object itself is required (in body) for this operation.
@@ -228,6 +230,8 @@ pub(super) struct ApiObjectBuilder<'a> {
     pub helper_module_prefix: &'a str,
     /// Operation ID, if any.
     pub op_id: Option<&'a str>,
+    /// Whether the operation is deprecated or not.
+    pub deprecated: bool,
     /// HTTP method for the operation - all builders (other than object builders)
     /// have this.
     pub method: Option<HttpMethod>,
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -34,7 +34,7 @@ test:
 	cargo test --all --all-features
 	# Compile the code generated through tests.
 	cd tests/test_pet && cargo check
-	cd tests/test_pet/cli && CARGO_TARGET_DIR=../target cargo check
+	cd tests/test_pet/cli && CARGO_TARGET_DIR=../target cargo build
```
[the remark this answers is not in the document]
wafflespeanut: Checking ensures that the code compiles and doesn't build anything unless required. We don't want the `test_pet` CLI because we're already building the huge `test_k8s` CLI (which I'd really like to get rid of, but not yet) - building binaries consume time and I don't want the CI to take long times.

src/v2/codegen/object.rs
```diff
@@ -56,6 +56,8 @@ pub struct OpRequirement {
     pub id: Option<String>,
     /// Description of this operation (if any), to be used for docs.
     pub description: Option<String>,
+
```
wafflespeanut: Since this is a public field, let's document to keep things consistent.

src/v2/codegen/object.rs
```diff
@@ -228,6 +230,8 @@ pub(super) struct ApiObjectBuilder<'a> {
     pub helper_module_prefix: &'a str,
     /// Operation ID, if any.
     pub op_id: Option<&'a str>,
+    /// Whether operation deprecated or not.
```
wafflespeanut: Whether *the* operation *is* deprecated or not.

wafflespeanut:
This looks good! Just a few nits and we can land this. :smile:

hindenbug:
@wafflespeanut Addressed the comments :)

wafflespeanut:

PR #132, opened 2019-10-10
Update Cratemeta values from CLI values

aeroaks:
update crate name and version from CLI args #130


wafflespeanut:
Nice work! Thanks for your PR! :smile:

PR #133, opened 2019-10-11
fix compilation errors: copy impl from trait and mark default

letmutx:
To fix compilation errors on the latest nightly.

### DIFF
--- a/core/src/v2/schema.rs
+++ b/core/src/v2/schema.rs
@@ -226,7 +226,13 @@ impl<T: TypedData> Apiv2Schema for T {
     }
 }
 
-impl<T> Apiv2Schema for Option<T> {}
+impl<T> Apiv2Schema for Option<T> {
+    default const NAME: Option<&'static str> = None;
+
+    default fn raw_schema() -> DefaultSchemaRaw {
+        Default::default()
+    }
+}
 
 impl<T: Apiv2Schema> Apiv2Schema for Option<T> {
     const NAME: Option<&'static str> = T::NAME;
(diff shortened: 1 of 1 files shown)


wafflespeanut:
Yeah, this makes sense, because we're implementing `Apiv2Schema` for both `Option<T: Apiv2Schema>` and `Option<T>`, so there's definitely an overlap.

core/src/v2/schema.rs
```diff
@@ -236,7 +244,16 @@ impl<T: Apiv2Schema> Apiv2Schema for Option<T> {
...
 
-impl<T, E> Apiv2Schema for Result<T, E> {}
+impl<T, E> Apiv2Schema for Result<T, E> {
+    default const NAME: Option<&'static str> = None;
+
+    /// Returns the raw schema for this object.
+    default fn raw_schema() -> DefaultSchemaRaw {
+        Default::default()
+    }
+
```
wafflespeanut: Please remove this newline.

core/src/v2/schema.rs
```diff
@@ -236,7 +244,16 @@ impl<T: Apiv2Schema> Apiv2Schema for Option<T> {
     }
 }
 
-impl<T, E> Apiv2Schema for Result<T, E> {}
+impl<T, E> Apiv2Schema for Result<T, E> {
+    default const NAME: Option<&'static str> = None;
+
+    /// Returns the raw schema for this object.
```
wafflespeanut: Can you also remove the doc comment here (and above)? IDEs and docs pages generally default to the documentation in trait methods. We should add doc comments here only if we're doing something different from what the actual doc comment says.

wafflespeanut:
Thanks for raising this PR! :smile:

PR #134, opened 2019-10-16
[#34] Add info field

### LINKED ISSUE #34: Use "info" field in spec
This is necessary for the CLI. We should use the [`info`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#infoObject) field to get the title, description, version, etc. (only if they exist - I don't think it's necessary to force that in CLI). An edge case would be that passing name and version through CLI args will override whatever name and version we obtained from the spec.

hindenbug:
fixes #34 
Here's what I came up with. Would appreciate any pointers to code/tests 

- [x] Add `Info` struct 
- [x]  Add `Contact` struct
- [x]  Add  `License` struct
- [x]  Fix `make test` build and test failures

### DIFF
--- a/src/v2/codegen/emitter.rs
+++ b/src/v2/codegen/emitter.rs
@@ -1,5 +1,6 @@
 use super::object::{ApiObject, ObjectField, OpRequirement, Parameter, Response};
 use super::state::{ChildModule, EmitterState};
+use super::CrateMeta;
 use crate::error::PaperClipError;
 use crate::v2::{
     im::ArcRwLock,
@@ -13,14 +14,13 @@ use crate::v2::{
 use failure::Error;
 use heck::{CamelCase, SnekCase};
 use itertools::Itertools;
-use url::Host;
-
 use std::collections::{BTreeSet, HashSet};
 use std::fmt::Debug;
 use std::fs;
 use std::ops::Deref;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
+use url::Host;
 
 /// Identifier used for `Any` generic parameters in struct definitions.
 pub(super) const ANY_GENERIC_PARAMETER: &str = "Any";
@@ -210,6 +210,15 @@ pub trait Emitter: Sized {
     fn generate(&self, api: &Api<Self::Definition>) -> Result<(), Error> {
         let state = self.state();
         state.reset_internal_fields();
+
+        let m = state.get_meta();
+
+        if m.borrow().is_none() {
+            let mut meta = CrateMeta::default();
+            meta.name = Some(api.info.title.clone());
+            meta.version = Some(api.info.version.clone());
+            state.set_meta(meta);
+        }
 
         // Add default coders.
         let mut coders = api.coders.clone();
--- a/src/v2/codegen/state.rs
+++ b/src/v2/codegen/state.rs
@@ -38,7 +38,7 @@ pub struct EmitterState {
     /// If crate metadata is specified, then `lib.rs` and `Cargo.toml` are generated
     /// along with the modules. This is gated behind `"cli"` feature.
     #[cfg(feature = "cli")]
-    crate_meta: Option<Rc<RefCell<CrateMeta>>>,
+    crate_meta: Rc<RefCell<Option<CrateMeta>>>,
 
     /* MARK: Internal fields that should be reset for each session. */
     /// Maps parent mod to immediate children. Used for declaring modules.
@@ -325,9 +325,9 @@ pub mod util {
             return Ok(());
         }
 
-        if let Some(m) = self.infer_crate_meta()? {
+        let cm = self.infer_crate_meta()?;
+        if let Some(meta) = cm.borrow().as_ref() {
             // Clap YAML
-            let meta = m.borrow();
             let clap_yaml = root.with_file_name("app.yaml");
             let base_content = template::render(
                 TEMPLATE::CLAP_YAML,
@@ -362,7 +362,9 @@ pub mod util {
     fn is_cli(&self) -> Result<bool, Error> {
         Ok(self
             .infer_crate_meta()?
-            .map(|m| m.borrow().is_cli)
+            .borrow()
+            .as_ref()
+            .map(|m| m.is_cli)
             .unwrap_or(false))
     }
 }
@@ -372,19 +374,23 @@ pub mod util {
 #[cfg(feature = "cli")]
 impl EmitterState {
     /// Sets the crate metadata for this session.
-    pub fn set_meta(&mut self, meta: CrateMeta) {
-        self.crate_meta = Some(Rc::new(RefCell::new(meta)));
+    pub fn set_meta(&self, meta: CrateMeta) {
+        *self.crate_meta.borrow_mut() = Some(meta)
+    }
+
+    pub(super) fn get_meta(&self) -> Rc<RefCell<Option<CrateMeta>>> {
+        self.crate_meta.clone()
     }
 
     /// Checks whether this session is for emitting a crate.
     fn is_crate(&self) -> bool {
-        self.crate_meta.is_some()
+        self.crate_meta.borrow().is_some()
     }
 
     /// Returns the path to the root module.
     fn root_module_path(&self) -> PathBuf {
-        if let Some(m) = self.crate_meta.as_ref() {
-            let meta = m.borrow();
+        let cm = self.crate_meta.borrow();
+        if let Some(meta) = cm.as_ref() {
             if meta.is_cli {
                 self.working_dir.join("main.rs")
             } else {
@@ -401,12 +407,13 @@ impl EmitterState {
         let is_cli = self.is_cli()?;
         man_path.set_file_name("Cargo.toml");
 
-        let m = match self.infer_crate_meta()? {
+        let cm = self.infer_crate_meta()?;
+        let m = cm.borrow();
+        let meta = match m.as_ref() {
             Some(c) => c,
             None => return Ok(()),
         };
 
-        let meta = m.borrow();
         if self.is_crate() {
             let contents = template::render(
                 TEMPLATE::CARGO_MANIFEST,
@@ -425,9 +432,9 @@ impl EmitterState {
     }
 
     /// Validates crate metadata, sets the unset fields and returns a reference.
-    fn infer_crate_meta(&self) -> Result<Option<Rc<RefCell<CrateMeta>>>, Error> {
-        if let Some(m) = self.crate_meta.as_ref() {
-            let mut meta = m.borrow_mut();
+    fn infer_crate_meta(&self) -> Result<Rc<RefCell<Option<CrateMeta>>>, Error> {
+        let mut cm = self.crate_meta.borrow_mut();
+        if let Some(meta) = &mut *cm {
             if meta.name.is_none() {
                 meta.name = Some(
                     fs::canonicalize(&self.working_dir)?
@@ -462,11 +469,16 @@ impl EmitterState {
 #[cfg(not(feature = "cli"))]
 impl EmitterState {
     /// This is a no-op.
-    pub fn set_meta(&mut self, _: CrateMeta) {}
+    pub fn set_meta(&self, _: CrateMeta) {}
+
+    /// This is a no-op
+    pub fn get_meta(&self) -> Rc<RefCell<Option<CrateMeta>>> {
+        Rc::new(RefCell::new(None))
+    }
 
     /// Always returns `Ok(None)`
-    fn infer_crate_meta(&self) -> Result<Option<Rc<RefCell<CrateMeta>>>, Error> {
-        Ok(None)
+    fn infer_crate_meta(&self) -> Result<Rc<RefCell<Option<CrateMeta>>>, Error> {
+        Ok(Rc::new(RefCell::new(None)))
     }
 
     /// Always returns the path to `mod.rs` in root.
@@ -510,7 +522,7 @@ impl Default for EmitterState {
             mod_prefix: "crate::",
             ns_sep: ".",
             #[cfg(feature = "cli")]
-            crate_meta: None,
+            crate_meta: Rc::new(RefCell::new(None)),
             base_url: RefCell::new("https://example.com".parse().expect("invalid URL?")),
             def_mods: RefCell::new(HashMap::new()),
             rel_paths: RefCell::new(HashSet::new()),
--- a/tests/test_app.rs
+++ b/tests/test_app.rs
@@ -75,6 +75,7 @@ fn test_simple_app() {
             check_json(
                 &mut resp,
                 json!({
+                  "info":{"title":"","version":""},
                   "definitions": {
                     "Pet": {
                       "properties": {
@@ -285,6 +286,7 @@ fn test_params() {
             check_json(
                 &mut resp,
                 json!({
+                  "info":{"title":"","version":""},
                   "definitions": {
                     "BadgeBody":{
                       "properties":{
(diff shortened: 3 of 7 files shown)


wafflespeanut:
Thanks for the PR! I've addressed the required changes in the issue.

src/v2/codegen/state.rs
```diff
@@ -372,8 +372,12 @@ pub mod util {
 #[cfg(feature = "cli")]
 impl EmitterState {
     /// Sets the crate metadata for this session.
-    pub fn set_meta(&mut self, meta: CrateMeta) {
-        self.crate_meta = Some(Rc::new(RefCell::new(meta)));
+    pub fn set_meta(&self, meta: CrateMeta) {
+        self.crate_meta = Rc::new(RefCell::new(Some(meta)));
```
hindenbug: `Question`:  Hi @wafflespeanut , this would change too, correct ?

src/v2/codegen/state.rs
```diff
@@ -372,8 +372,12 @@ pub mod util {
...
+    }
+
+    pub(super) fn get_meta(&self) -> Option<Rc<RefCell<Option<CrateMeta>>>> {
+        Some(self.crate_meta.clone())
```
hindenbug: Same about this 🤔

src/v2/codegen/state.rs
```diff
@@ -510,7 +519,7 @@ impl Default for EmitterState {
             mod_prefix: "crate::",
             ns_sep: ".",
             #[cfg(feature = "cli")]
-            crate_meta: None,
+            crate_meta: Rc::new(RefCell::new(None)),
```
hindenbug: ☝️ too

src/v2/codegen/state.rs
wafflespeanut: That is correct.

src/v2/codegen/state.rs
wafflespeanut: Yes, everything involving `crate_meta` field. :smile:

src/v2/codegen/state.rs
wafflespeanut: Yep!

hindenbug:
Hi @wafflespeanut am running into some more issues with the borrow,  :D  Need some help yet again :)

I am actually confused about these, 
https://github.com/wafflespeanut/paperclip/blob/4bb46bd853a2a01e9882a0da3e03ad5c6cb84db2/src/v2/codegen/state.rs#L428
https://github.com/wafflespeanut/paperclip/blob/4bb46bd853a2a01e9882a0da3e03ad5c6cb84db2/src/v2/codegen/state.rs#L468

and 

```
argo clippy --all -- -D clippy::all
    Checking paperclip v0.3.0 (/Users/manoj/OSS/rust/paperclip)
error[E0609]: no field `name` on type `std::cell::Ref<'_, std::option::Option<v2::codegen::CrateMeta>>`
   --> src/v2/codegen/state.rs:335:32
    |
335 |                     name: meta.name.as_ref().unwrap(),
    |                                ^^^^ unknown field

error[E0609]: no field `version` on type `std::cell::Ref<'_, std::option::Option<v2::codegen::CrateMeta>>`
   --> src/v2/codegen/state.rs:336:52
    |
336 |                     version: &format!("{:?}", meta.version.as_ref().unwrap()),
    |                                                    ^^^^^^^ unknown field

error[E0609]: no field `is_cli` on type `std::cell::Ref<'_, std::option::Option<v2::codegen::CrateMeta>>`
   --> src/v2/codegen/state.rs:365:33
    |
365 |             .map(|m| m.borrow().is_cli)
    |                                 ^^^^^^ unknown field
```

wafflespeanut:
So, the changes depend on the types. Previously it was `Option<Rc<RefCell<CrateMeta>>>`, but now it's `Rc<RefCell<Option<CrateMeta>>>`. So, `infer_crate_meta` function's return type now changes to the new type. And currently, we do...

```rust
// deal with option first followed by refcell borrow
if let Some(foo) = /* crate meta */ {
    let meta = m.borrow(); // or borrow_mut();
    ...
}
```

But now, this will become:

```rust
let borrow = /* refcell borrow first */;
// then go for option deref
if let Some(m) = borrow {
    ...
}
```

Does this make sense?

src/v2/codegen/state.rs
```diff
@@ -372,18 +372,22 @@ pub mod util {
...
 impl EmitterState {
     /// Sets the crate metadata for this session.
-    pub fn set_meta(&mut self, meta: CrateMeta) {
-        self.crate_meta = Some(Rc::new(RefCell::new(meta)));
+    pub fn set_meta(&self, meta: CrateMeta) {
+        self.crate_meta = Rc::new(RefCell::new(Some(meta)));
+    }
+
+    pub(super) fn get_meta(&self) -> Option<Rc<RefCell<Option<CrateMeta>>>> {
+        Some(self.crate_meta.clone())
```
wafflespeanut: This could just be `Rc<RefCell<Option<CrateMeta>>>`

hindenbug:
> So, the changes depend on the types. Previously it was `Option<Rc<RefCell<CrateMeta>>>`, but now it's `Rc<RefCell<Option<CrateMeta>>>`. So, `infer_crate_meta` function's return type now changes to the new type. And currently, we do...
> 
> ```rust
> // deal with option first followed by refcell borrow
> if let Some(foo) = /* crate meta */ {
>     let meta = m.borrow(); // or borrow_mut();
>     ...
> }
> ```
> 
> But now, this will become:
> 
> ```rust
> let borrow = /* refcell borrow first */;
> // then go for option deref
> if let Some(m) = borrow {
>     ...
> }
> ```
> 
> Does this make sense?

@wafflespeanut so this means the function's return type would NOT be a `Result` ? 🤔

hindenbug:
So I attempted again, and now ending up with 👇  - @wafflespeanut 

```rust
error[E0594]: cannot assign to `self.crate_meta` which is behind a `&` reference
   --> src/v2/codegen/state.rs:377:9
    |
376 |     pub fn set_meta(&self, meta: CrateMeta) {
    |                     ----- help: consider changing this to be a mutable reference: `&mut self`
377 |         self.crate_meta = Rc::new(RefCell::new(Some(meta)));
    |         ^^^^^^^^^^^^^^^ `self` is a `&` reference, so the data it refers to cannot be written

error[E0594]: cannot assign to `self.crate_meta` which is behind a `&` reference
   --> src/v2/codegen/state.rs:377:9
    |
376 |     pub fn set_meta(&self, meta: CrateMeta) {
    |                     ----- help: consider changing this to be a mutable reference: `&mut self`
377 |         self.crate_meta = Rc::new(RefCell::new(Some(meta)));
    |         ^^^^^^^^^^^^^^^ `self` is a `&` reference, so the data it refers to cannot be written

error[E0596]: cannot borrow `*m` as mutable, as it is behind a `&` reference
   --> src/v2/codegen/state.rs:437:24
    |
436 |         if let Some(m) = cm.as_ref() {
    |                     - help: consider changing this to be a mutable reference: `&mut v2::codegen::CrateMeta`
437 |             let meta = &mut *m;
    |                        ^^^^^^^ `m` is a `&` reference, so the data it refers to cannot be borrowed as mutable

error[E0596]: cannot borrow `*m` as mutable, as it is behind a `&` reference
   --> src/v2/codegen/state.rs:437:24
    |
436 |         if let Some(m) = cm.as_ref() {
    |                     - help: consider changing this to be a mutable reference: `&mut v2::codegen::CrateMeta`
437 |             let meta = &mut *m;
    |                        ^^^^^^^ `m` is a `&` reference, so the data it refers to cannot be borrowed as mutable
```

src/v2/codegen/state.rs
```diff
@@ -372,19 +373,23 @@ pub mod util {
 #[cfg(feature = "cli")]
 impl EmitterState {
     /// Sets the crate metadata for this session.
-    pub fn set_meta(&mut self, meta: CrateMeta) {
-        self.crate_meta = Some(Rc::new(RefCell::new(meta)));
+    pub fn set_meta(&self, meta: CrateMeta) {
+        self.crate_meta = Rc::new(RefCell::new(Some(meta)));
```
wafflespeanut: We should utilize refcell's interior mutability (`RefCell::borrow_mut`) here, right? :smile:

src/v2/codegen/state.rs
```diff
@@ -425,9 +431,10 @@ impl EmitterState {
     }
 
     /// Validates crate metadata, sets the unset fields and returns a reference.
-    fn infer_crate_meta(&self) -> Result<Option<Rc<RefCell<CrateMeta>>>, Error> {
-        if let Some(m) = self.crate_meta.as_ref() {
-            let mut meta = m.borrow_mut();
+    fn infer_crate_meta(&self) -> Result<Option<Rc<RefCell<Option<CrateMeta>>>>, Error> {
```
wafflespeanut: This function should return `Result<Rc<RefCell<Option<CrateMeta>>>, Error>` - is there any particular reason for using `Result<Option<...>, Error>`?

src/v2/codegen/state.rs
```diff
@@ -362,7 +363,7 @@ pub mod util {
     fn is_cli(&self) -> Result<bool, Error> {
         Ok(self
             .infer_crate_meta()?
-            .map(|m| m.borrow().is_cli)
+            .map(|m| m.borrow().as_ref().unwrap().is_cli)
```
wafflespeanut: This will change once `infer_crate_meta` signature changes below.

src/v2/codegen/emitter.rs
```diff
@@ -211,6 +211,18 @@ pub trait Emitter: Sized {
         let state = self.state();
         state.reset_internal_fields();
 
+        let m = state.get_meta();
+        if let Some(mut meta) = m.borrow_mut().clone() {
```
wafflespeanut: This looks like we're setting title and version only if crate meta is already set. We should do it the other way. If crate meta isn't set, then we should create one from the API, since we know that the API always has title and version?

src/v2/codegen/emitter.rs
```diff
@@ -211,6 +211,18 @@ pub trait Emitter: Sized {
...
+            if meta.name.is_none() {
+                meta.name = Some(api.clone().info.title)
+            }
+            if meta.version.is_none() {
+                meta.version = Some(api.clone().info.version)
```
wafflespeanut: What's wrong with doing `api.info.version.clone()`? We don't have to clone the entire `api` (it's huge).

src/v2/codegen/emitter.rs
hindenbug: Thanks @wafflespeanut for the feedback :) I have a few question on ☝️ 
1) Do you mean we need to check `state.crate_meta.borrow().is_none` by making `crate_meta` public ?
2) Will this check be using `cli` feature flag ?  🤔

src/v2/codegen/state.rs
hindenbug: Not sure if I got it right? 🤔  You mean `self.crate_meta.borrow_mut() = Some(meta);`

wafflespeanut: `borrow_mut` produces a temporary borrow which can be dereferenced to the mutable version of meta. So,
```rust
*self.crate_meta.borrow_mut() = Some(meta);
```

src/v2/codegen/emitter.rs
wafflespeanut: We have `get_meta` and `set_meta` methods for getting/setting. If `cli` isn't enabled, then whatever we do will be no-op, so you shouldn't have to worry about `cli` flag.

src/v2/codegen/emitter.rs
```diff
@@ -211,6 +211,22 @@ pub trait Emitter: Sized {
         let state = self.state();
         state.reset_internal_fields();
 
+        let m = state.get_meta();
+        let borrowed_cm = m.borrow();
+
+        if borrowed_cm.is_none() {
+            if let Some(meta) = &mut *m.borrow_mut() {
```
wafflespeanut: The first problem is that `borrowed_cm` is already borrowed immutably, so a mutable borrow will cause a panic at runtime (in this case, during tests). We can get rid of `borrowed_cm` entirely, since we're not using beyond it the `if` condition.

The other problem is that if we've already entered the block, it means that crate meta is `None`, so we won't ever satisfy the condition which is checking for `Some` pattern. What we should do is if it's `None`, then we should create a `CrateMeta`, set the name and version, and call `set_meta`.

src/v2/codegen/state.rs
```diff
@@ -325,9 +325,10 @@ pub mod util {
             return Ok(());
         }
 
-        if let Some(m) = self.infer_crate_meta()? {
+        if let m = self.infer_crate_meta()? {
```
wafflespeanut: This doesn't seem to match any pattern. We should bind the output to some variable and then match it.

```rust
let meta = self.infer_crate_meta()?;
if let Some(m) = meta.borrow() {
    ...
}
```

src/v2/codegen/emitter.rs
```diff
@@ -211,6 +211,15 @@ pub trait Emitter: Sized {
         let state = self.state();
         state.reset_internal_fields();
 
+        let m = state.get_meta();
+
+        if m.borrow_mut().is_none() {
```
wafflespeanut: A readable reference (`.borrow()`) is fine here?

src/v2/codegen/state.rs
```diff
@@ -400,13 +406,13 @@ impl EmitterState {
         let mut man_path = self.root_module_path();
         let is_cli = self.is_cli()?;
         man_path.set_file_name("Cargo.toml");
+        let cm = self.crate_meta.borrow();
```
wafflespeanut: This should be `self.infer_crate_meta()?`, right?

tests/test_app.rs
```diff
@@ -93,7 +94,10 @@ fn test_simple_app() {
                           "format": "date-time",
                           "type": "string"
                         },
-                        "uid": {}
+                        "uid": {
```
wafflespeanut: Please ignore this. This is my bad. I'm working on it.

hindenbug:
@wafflespeanut not sure whats wrong, but am getting different test failures than on the ci 🤔

wafflespeanut:
@hindenbug Please post the errors? If it's about mismatching JSON, then I'm guessing it's because of UUID, in which case you can delete `Cargo.lock` (in root) and try it again.

hindenbug:
@wafflespeanut Thanks, removing `Cargo.lock` helps.

hindenbug:
getting this strange error, what could I be doing wrong here 🤔 
```rust
---- test_root_mod stdout ----
thread 'test_root_mod' panicked at 'missing file: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/libcore/result.rs:1165:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

failures:
    test_root_mod

test result: FAILED. 17 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '-p paperclip --test test_k8s'
The command "cargo test --all --all-features" exited with 101.
$ cd tests/test_k8s && cargo check
error: failed to parse manifest at `/home/travis/build/wafflespeanut/paperclip/tests/test_k8s/Cargo.toml`

Caused by:
  Error parsing major identifier for key `package.version`
The command "cd tests/test_k8s && cargo check" exited with 101.
```

wafflespeanut:
Okay, looks like that's going beyond the scope of this PR, for which I may have to do a workaround for now. I'll take care of this. Thank you for all your hard work! :smile:

hindenbug:
Thanks @wafflespeanut for your patience and guidance :smiley:

PR #139, opened 2019-11-14
Type annotations in actix plugin

sfisol:
These annotations allowed me to compile the crate and use it in my actix-based project.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ log = "0.4"
 paperclip-actix = { path = "plugins/actix-web", version = "0.1.0", optional = true }
 paperclip-core = { path = "core", version = "0.1.0" }
 paperclip-macros = { path = "macros", version = "0.2.0", optional = true }
-parking_lot = { version = "0.8" }
+parking_lot = { version = "0.10" }
 regex = { version = "1.1", optional = true }
 reqwest = { version = "0.9", optional = true }
 # rustfmt-nightly = { version = "1.2.2", optional = true }
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -20,7 +20,7 @@ lazy_static = "1.3"
 log = "0.4"
 mime = { git = "https://github.com/hyperium/mime", features = ["serde1"] }
 paperclip-macros = { path = "../macros", version = "0.2.0" }
-parking_lot = "0.9"
+parking_lot = "0.10"
 regex = "1.1"
 serde = { version = "1.0", features = ["derive", "rc"] }
 serde_json = "1.0"
--- a/plugins/actix-web/Cargo.toml
+++ b/plugins/actix-web/Cargo.toml
@@ -16,4 +16,4 @@ actix-web = "1.0.4"
 lazy_static = "1.3"
 paperclip-core = { path = "../../core", version = "0.1.0", features = ["actix"] }
 paperclip-macros = { path = "../../macros", version = "0.2.0", features = ["actix"] }
-parking_lot = "0.8"
+parking_lot = "0.10"


wafflespeanut:
I'm okay with the changes, but I'm really curious as to what error you've encountered and why it cropped up in the `codegen` module (which is not used by the actix plugin). One other reason is that `fmt::Result` is [a type alias](https://doc.rust-lang.org/nightly/std/fmt/type.Result.html) to the `Result` type, which is already imported in prelude automatically by rustc, so this shouldn't have been a problem at all. :thinking:

sfisol:
When I add original paperclip dependency as:
`paperclip = { path = "../paperclip", features = ["actix"] }`

Then it gives me errors:
```
error[E0282]: type annotations needed
   --> .../paperclip/src/v2/codegen/impls.rs:247:13
    |
247 | /             builder.struct_fields_iter().try_for_each(|field| {
248 | |                 if field.prop.is_required() {
249 | |                     f.write_str("\n            ")?;
250 | |                     if field.prop.is_parameter() {
...   |
264 | |                 Ok(())
265 | |             })?;
    | |_______________^ cannot infer type

error[E0282]: type annotations needed
   --> .../paperclip/src/v2/codegen/object.rs:731:9
    |
731 | /         self.struct_fields_iter().try_for_each(|field| {
732 | |             let (cc, sk) = (field.name.to_camel_case(), field.name.to_snek_case());
733 | |             if needs_container {
734 | |                 self.write_parameter_if_required(
...   |
759 | |             Ok(())
760 | |         })?;
    | |___________^ cannot infer type

error[E0282]: type annotations needed
   --> .../paperclip/src/v2/codegen/object.rs:791:9
    |
791 | /         self.fields.iter().try_for_each(|field| {
792 | |             let mut new_name = field.name.to_snek_case();
793 | |             // Check if the field matches a Rust keyword and add '_' suffix.
794 | |             if RUST_KEYWORDS.iter().any(|&k| k == new_name) {
...   |
836 | |             Ok(())
837 | |         })?;
    | |___________^ cannot infer type

error: aborting due to 3 previous errors
```

Maybe another dependency is messing things up? Others are:
```
actix = "0.8"
actix-web = { version="1.0.8", features=["ssl"] }
chrono = "0.4"
lazy_static = "1.4"
futures = "0.1"
rand = "0.7"
roxmltree = "0.7"
rust_decimal = "1.0"
serde = { version = "1.0", features = ["derive"] }
serde_json = "1.0"
```

I can prepare an example with minimal amount of deps if you want.

wafflespeanut:
Which version of Rust are you using? `codegen` feature isn't required for the actix plugin, so we could try enabling `v2` and disabling the default features. Could you try this in your manifest?

```toml
paperclip = { path = "../paperclip", features = ["actix", "v2"], default-features = false }
```

sfisol:
Invoking "rustc --version" in paperclip directory gives me: rustc 1.40.0-nightly (4a8c5b20c 2019-10-23)
Invoking this in my project gives me: rustc 1.40.0-nightly (10a52c25c 2019-10-24)
Adding "deafult-features = false" makes paperclip compile successful (no matter whether I try with "v2" or without)
I noticed on other project that just adding paperclip dependency with actix feature does not render errors. I'll prepare some minimal example so we could see if it is a matter of some third-party deps or some special paperclip usage.

sfisol:
I've found it. It's "ssl" feature added to actix-web. There can be no code in main, and Cargo.toml:

```
[package]
name = "paperexample"
version = "0.1.0"

[dependencies]
actix-web = { version="1.0.4", features=["ssl"] }
paperclip = { git = "https://github.com/wafflespeanut/paperclip", features = ["actix"] }
```

...and I get compilation errors. However disabling codegen is sufficient workaround for me now.

wafflespeanut:
Okay. Do you still want this change to land in tree?

sfisol:
I don't need it anymore, but others can possibly run into the same trouble. The question is if these annotations are fixing the problem or just workarounding. But I'm not such an expert in Rust to track down the root cause of this, whether this is something in actix-web or actix-net or tokio-openssl..

Issue #115, opened 2019-10-02
Add rest of the fields in root


wafflespeanut:
We have a [`GenericApi<S>` struct](https://github.com/wafflespeanut/paperclip/blob/e5fc2a0e5eba052eb594fca378caedad3c7a692a/core/src/v2/models.rs#L86) to represent swagger files. It's missing a number of fields as of now. Here's the [spec for list of fields](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#swagger-object). The struct needs the following fields with appropriate types:

Field | Optional | Type
--- | --- | ---
`parameters` | Yes | BTreeMap
`responses` | Yes | BTreeMap
`security_definitions` | Yes | BTreeMap (with new struct)
`security` | Yes | Vec (with new struct)
`tags` | Yes | Vec (with new struct)
`external_docs` | Yes | New struct

Feel free to ask any questions!

rootbid:
Hey @wafflespeanut, Let me get my feet wet with this.

Question: [Info Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#infoObject) for the `info` field have 2 child objects [Contact](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#contactObject) and [License](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#licenseObject). Does it make sense to create individual structs for them?

wafflespeanut:
I'll assign it to you! And yes, you're right about creating new structs. :smile:

rootbid:
Things have been on/off for me, if someone else is interested they might give it a try.

wafflespeanut:
No worries! Thanks for letting me know! :smile:

aeroaks:
Hi, I would like to pick this up. :)

wafflespeanut:
You can have it! :smile:

aeroaks:
Hi, I am looking into the `GenericApi` Struct and want to have some example/explanation where and how I can define new `structs` to be used for defining the missing fields.

wafflespeanut:
@aeroaks Any news on this? Do you need help?

aeroaks:
@wafflespeanut I was busy with some work and home stuff. I can start working on this from tomorrow, I will contact you when I feel stuck. If that is fine then I would like to continue!

wafflespeanut:
No problem!

aeroaks:
[...]

PR #140, opened 2019-11-18
Add rest of the fields in root #115

aeroaks:
first attempt at adding fields

### DIFF
--- a/core/src/v2/models.rs
+++ b/core/src/v2/models.rs
@@ -103,11 +103,17 @@ pub struct Api<P, R, S> {
     pub produces: BTreeSet<MediaRange>,
     #[serde(default, skip_serializing_if = "BTreeSet::is_empty")]
     pub schemes: BTreeSet<OperationProtocol>,
-    pub parameters: BTreeMap<String, Parameter<S>>,
-    pub responses: BTreeMap<String, Response<S>>,
-    pub security_definitions: BTreeMap<String, SecurityDefinition<S>>
-    pub security: Vec<Security>,
+    #[serde(default = "BTreeMap::new", skip_serializing_if = "BTreeMap::is_empty")]
+    pub parameters: BTreeMap<String, P>,
+    #[serde(default = "BTreeMap::new", skip_serializing_if = "BTreeMap::is_empty")]
+    pub responses: BTreeMap<String, R>,
+    #[serde(default, rename = "securityDefinitions")]
+    pub security_definitions: BTreeMap<String, SecurityScheme>
+    #[serde(skip_serializing_if = "BTreeSet::is_empty")]
+    pub security: BTreeMap<String, Vec<String>>,
+    #[serde(skip_serializing_if = "Vec::is_none")]
     pub tags: Vec<Tags>,
+    #[serde(default, rename = "externalDocs")]
     pub external_docs: Option<ExternalDocs>,
     #[serde(
         default,
@@ -223,40 +229,25 @@ pub struct License {
     pub url: Option<String>,
 }
 
-/// Security Definition object.
-///
-/// https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityDefinitionsObject
-#[derive(Clone, Debug, Default, Serialize, Deserialize)]
-pub struct SecurityDefinition {
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub name: Option<SecurityScheme>,
-}
-
 /// Security Scheme object.
 ///
 /// https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#security-scheme-object
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
 pub struct SecurityScheme {
-    pub type: String,
+    #[serde(rename = "type")]
+    pub type_: String,
     pub name: String,
-    pub in: String,
+    #[serde(rename = "in")]
+    pub in_: String,
     pub flow: String,
-    pub authorizationUrl: String,
-    pub tokenUrl: String,
-    pub scopes: Scopes,
+    #[serde(rename = "authorizationUrl")]
+    pub authorization_url: String,
+    #[serde(rename = "tokenUrl")]
+    pub token_url: String,
+    pub scopes: BTreeMap<String, String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub description: Option<String>,
 }
-
-/// Scopes object.
-///
-/// https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#scopesObject
-#[derive(Clone, Debug, Default, Serialize, Deserialize)]
-pub struct Scopes {
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub name: Option<String>,
-}
-
 
 /// Tag object.
 ///
@@ -267,7 +258,8 @@ pub struct Tag {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub description: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub externalDocs: Option<ExternalDocs>,
+    #[serde(rename = "externalDocs")]
+    pub external_docs: Option<ExternalDocs>,
 }
 
 /// External Documentation object.
@@ -279,16 +271,6 @@ pub struct ExternalDocs {
     pub description: Option<String>,
     pub url: String,
 }
-
-/// Security object.
-///
-/// https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityRequirementObject
-#[derive(Clone, Debug, Default, Serialize, Deserialize)]
-pub struct Security {
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub name: Option<String>,
-}
-
 
 /// Path item that can be traversed and resolved for codegen.
 pub type ResolvablePathItem<S> = PathItem<ResolvableParameter<S>, ResolvableResponse<S>>;
--- a/core/src/v2/models.rs
+++ b/core/src/v2/models.rs
@@ -110,7 +110,8 @@ pub struct Api<P, R, S> {
     #[serde(
         default, 
         rename = "securityDefinitions", 
-        skip_serializing_if = "BTreeMap::is_empty")]
+        skip_serializing_if = "BTreeMap::is_empty"
+    )]
     pub security_definitions: BTreeMap<String, SecurityScheme>,
     #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
     pub security: BTreeMap<String, Vec<String>>,
--- a/core/src/v2/models.rs
+++ b/core/src/v2/models.rs
@@ -107,13 +107,13 @@ pub struct Api<P, R, S> {
     pub parameters: BTreeMap<String, P>,
     #[serde(default = "BTreeMap::new", skip_serializing_if = "BTreeMap::is_empty")]
     pub responses: BTreeMap<String, R>,
-    #[serde(default, rename = "securityDefinitions")]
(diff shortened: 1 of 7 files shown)


core/src/v2/models.rs
```diff
@@ -221,6 +223,73 @@ pub struct License {
     pub url: Option<String>,
 }
 
+/// Security Definition object.
+///
+/// https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityDefinitionsObject
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct SecurityDefinition {
```
wafflespeanut: This struct is not necessary, as this is actually in the BTreeMap.

core/src/v2/models.rs
```diff
@@ -221,6 +223,73 @@ pub struct License {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub name: Option<SecurityScheme>,
+}
+
+/// Security Scheme object.
+///
+/// https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#security-scheme-object
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct SecurityScheme {
+    pub type: String,
```
wafflespeanut: This should be `type_` (as it interferes with Rust keyword) and it should have `#[serde(rename = "type")]`

core/src/v2/models.rs
```diff
@@ -221,6 +223,73 @@ pub struct License {
...
+    pub name: String,
+    pub in: String,
```
wafflespeanut: Similar to type, this should be `in_` with a serde rename.

core/src/v2/models.rs
```diff
@@ -221,6 +223,73 @@ pub struct License {
...
+    pub flow: String,
+    pub authorizationUrl: String,
+    pub tokenUrl: String,
```
wafflespeanut: These fields should be snake cased (`authorization_url` and `token_url`) with serde renames.

core/src/v2/models.rs
```diff
@@ -221,6 +223,73 @@ pub struct License {
...
+    pub scopes: Scopes,
```
wafflespeanut: This should be `BTreeMap<String, String>`

core/src/v2/models.rs
```diff
@@ -221,6 +223,73 @@ pub struct License {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub description: Option<String>,
+}
+
+/// Scopes object.
+///
+/// https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#scopesObject
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct Scopes {
```
wafflespeanut: We can get rid of this struct too.

core/src/v2/models.rs
```diff
@@ -103,10 +103,12 @@ pub struct Api<P, R, S> {
...
     #[serde(default, skip_serializing_if = "BTreeSet::is_empty")]
     pub schemes: BTreeSet<OperationProtocol>,
-    #[serde(default = "BTreeMap::new", skip_serializing_if = "BTreeMap::is_empty")]
-    pub parameters: BTreeMap<String, P>,
-    #[serde(default = "BTreeMap::new", skip_serializing_if = "BTreeMap::is_empty")]
-    pub responses: BTreeMap<String, R>,
+    pub parameters: BTreeMap<String, Parameter<S>>,
+    pub responses: BTreeMap<String, Response<S>>,
+    pub security_definitions: BTreeMap<String, SecurityDefinition<S>>
+    pub security: Vec<Security>,
```
wafflespeanut: This should be `BTreeMap<String, Vec<String>>`

core/src/v2/models.rs
```diff
@@ -221,6 +223,73 @@ pub struct License {
...
+/// Tag object.
+///
+/// https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#tag-object
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct Tag {
+    pub name: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub description: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub externalDocs: Option<ExternalDocs>,
```
wafflespeanut: Snake case field name with serde rename.

core/src/v2/models.rs
```diff
@@ -221,6 +223,73 @@ pub struct License {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub description: Option<String>,
+    pub url: String,
+}
+
+/// Security object.
+///
+/// https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityRequirementObject
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct Security {
```
wafflespeanut: We can get rid of this struct too.

core/src/v2/models.rs
aeroaks: Interesting, so `#[serde(rename = "type")]` chagnes it during de/serialisation?

core/src/v2/models.rs
aeroaks: Could you please explain what `BTreeMap<String, String>` is doing?

wafflespeanut: It's a collection of strings mapped to strings. That's exactly what the [scopes object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#scopesObject) demands. A field pattern `{name}` with type `string` means that for each string, there's a value of type `string`, which is strings mapped to strings.

core/src/v2/models.rs
```diff
@@ -103,10 +103,12 @@ pub struct Api<P, R, S> {
...
-    #[serde(default = "BTreeMap::new", skip_serializing_if = "BTreeMap::is_empty")]
-    pub parameters: BTreeMap<String, P>,
-    #[serde(default = "BTreeMap::new", skip_serializing_if = "BTreeMap::is_empty")]
-    pub responses: BTreeMap<String, R>,
+    pub parameters: BTreeMap<String, Parameter<S>>,
+    pub responses: BTreeMap<String, Response<S>>,
+    pub security_definitions: BTreeMap<String, SecurityDefinition<S>>
+    pub security: Vec<Security>,
+    pub tags: Vec<Tags>,
+    pub external_docs: Option<ExternalDocs>,
```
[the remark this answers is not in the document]
wafflespeanut: Sorry, this wasn't default, this should be `#[serde(rename = "externalDocs", skip_serializing_if = "Option::is_none")]`

core/src/v2/models.rs
```diff
@@ -107,6 +107,14 @@ pub struct Api<P, R, S> {
     pub parameters: BTreeMap<String, P>,
     #[serde(default = "BTreeMap::new", skip_serializing_if = "BTreeMap::is_empty")]
     pub responses: BTreeMap<String, R>,
+    #[serde(default, rename = "securityDefinitions")]
+    pub security_definitions: BTreeMap<String, SecurityScheme>
+    #[serde(skip_serializing_if = "BTreeSet::is_empty")]
+    pub security: BTreeMap<String, Vec<String>>,
+    #[serde(skip_serializing_if = "Vec::is_none")]
```
wafflespeanut: This should be `#[serde(default, skip_serializing_if = "Vec::is_empty")]`.

core/src/v2/models.rs
```diff
@@ -107,6 +107,14 @@ pub struct Api<P, R, S> {
     pub parameters: BTreeMap<String, P>,
     #[serde(default = "BTreeMap::new", skip_serializing_if = "BTreeMap::is_empty")]
     pub responses: BTreeMap<String, R>,
+    #[serde(default, rename = "securityDefinitions")]
+    pub security_definitions: BTreeMap<String, SecurityScheme>
```
wafflespeanut: This should also have a `skip_serializing_if = "BTreeMap::is_empty"`

core/src/v2/models.rs
```diff
@@ -107,6 +107,14 @@ pub struct Api<P, R, S> {
...
+    #[serde(skip_serializing_if = "BTreeSet::is_empty")]
```
wafflespeanut: This should also have a `default`

aeroaks:
I do not understand why the build is failing.

wafflespeanut:
Because a field is missing comma. I've updated the review comment.

aeroaks:
Hi, Could you please help me with the `test` part in the failing build?

core/src/v2/models.rs
```diff
@@ -107,6 +107,18 @@ pub struct Api<P, R, S> {
...
     pub responses: BTreeMap<String, R>,
+    #[serde(
+        default,
+        rename = "securityDefinitions",
+        skip_serializing_if = "BTreeMap::is_empty"
+    )]
+    pub security_definitions: BTreeMap<String, SecurityScheme>,
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub security: BTreeMap<String, Vec<String>>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
```
wafflespeanut: Please add `default` to serde attributes. That's what's causing the build failure.

core/src/v2/models.rs
```diff
@@ -107,6 +107,18 @@ pub struct Api<P, R, S> {
...
     #[serde(default = "BTreeMap::new", skip_serializing_if = "BTreeMap::is_empty")]
     pub responses: BTreeMap<String, R>,
+    #[serde(
+        default,
+        rename = "securityDefinitions",
+        skip_serializing_if = "BTreeMap::is_empty"
+    )]
+    pub security_definitions: BTreeMap<String, SecurityScheme>,
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub security: BTreeMap<String, Vec<String>>,
```
wafflespeanut: One last thing I've overlooked apparently. This should be `Vec<BTreeMap<String, Vec<String>>>`.

core/src/v2/models.rs
```diff
@@ -107,6 +107,18 @@ pub struct Api<P, R, S> {
     pub parameters: BTreeMap<String, P>,
     #[serde(default = "BTreeMap::new", skip_serializing_if = "BTreeMap::is_empty")]
     pub responses: BTreeMap<String, R>,
+    #[serde(
+        default,
+        rename = "securityDefinitions",
+        skip_serializing_if = "BTreeMap::is_empty"
+    )]
+    pub security_definitions: BTreeMap<String, SecurityScheme>,
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
```
wafflespeanut: The method should now be `Vec::is_empty`

aeroaks: Is there a way to run these changes locally first before push to github?

core/src/v2/models.rs
```diff
@@ -221,6 +233,49 @@ pub struct License {
     pub url: Option<String>,
 }
 
+/// Security Scheme object.
+///
+/// https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#security-scheme-object
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct SecurityScheme {
```
wafflespeanut: Can you please make all the fields in this struct optional? Some of the fields are required only for specific schemes. Since they depend on the scheme, I'll have to enforce it elsewhere in code.

core/src/v2/models.rs
wafflespeanut: I somehow missed this. Yes, you can use `make` to build and test the repository. Please check the `Makefile` in the repo root for building and testing commands.

aeroaks:
Tests are failing in the CI build. Could you please explain what needs to be done to solve this?

PR #153, opened 2020-02-13
Decimal field type (as a feature)

sfisol:
I've found this feature useful for me - works perfect with actix plugin.


wafflespeanut:
Awesome! Thanks for the PR! :smile:

PR #154, opened 2020-02-16
Updated actix-web to 2.0, updated futures to 0.3

sfisol:
I've updated versions of all actix libs and futures lib.

It worked for one of my project but I'm not sure if I've done all this properly. I'll have to do more testing (two another projects awaits), but I'm creating this PR as a draft so someone else can take a look at it to maybe spot something.

As far as i noticed this won't compile with actix-web openssl feature (happily it works with rustls)
Also I couldn't manage to wrap async/await handler with api_v2_operation macro, I'll investigate it shortly.

### DIFF
--- a/tests/test_app.rs
+++ b/tests/test_app.rs
@@ -10,7 +10,7 @@ use actix_service::ServiceFactory;
 use actix_web::dev::{MessageBody, Payload, ServiceRequest, ServiceResponse};
 use actix_web::{App, Error, FromRequest, HttpRequest, HttpServer, Responder};
 use chrono;
-use futures::future::{Future, Ready, ok as fut_ok, ready};
+use futures::future::{ok as fut_ok, ready, Future, Ready};
 use paperclip::actix::{api_v2_operation, api_v2_schema, web, OpenApiExt};
 use parking_lot::Mutex;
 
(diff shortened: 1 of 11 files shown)


wafflespeanut:
Thank you for working on this! Let me know if something's needed from my end! :sweat_smile:

sfisol:
It looks like in new actix-web there are only handlers returning futures, so we do not need to double-implement Apiv2Operation for Future-or-not return values.

So I only implemented it for regular Future with Output being OperationModifier. Future itself do not need to implement OperationModifier. Then we do not need FutureWrapper anymore.

I've tried to implement documenting responded for impl Responder but I've failed. I think this didn't work earlier either?

In emit_v2_operation when handler returns Responder we need to check if the function is async. If yes, then we just wrap Responder as usual. If it is not async, we can wrap Responder wrapper with additional impl Future, so we do not need even to implement Future for ResponderWrapper - and it allow us to stay with single implementations for Apiv2Operation.

I know my description may be a little unintelligible, so maybe just look at the changes and say if there is anything wrong or need clarification :-)

In near future I'll maybe try to implement documenting error codes somehow... or at least allow to pass it manually when decorating handler.

wafflespeanut:
So I only implemented it for regular Future with Output being OperationModifier. Future itself do not need to implement OperationModifier. Then we do not need FutureWrapper anymore.

Awesome!

> I've tried to implement documenting responses for impl Responder but I've failed. I think this didn't work earlier anyway?

No, it didn't work earlier. All we *can* do is return an empty schema for `impl Responder` as (I believe) it's impossible to get its concrete type in proc macros.

> In emit_v2_operation when handler returns Responder we need to check if the function is async. If yes, then we just wrap Responder as usual. If it is not async, we can wrap Responder wrapper with additional impl Future, so we do not need even to implement Future for ResponderWrapper - and it allow us to stay with single implementations for Apiv2Operation.

This makes sense.

> In near future I'll maybe try to implement documenting error codes somehow... or at least allow to pass it manually when decorating handler.

Looking forward to this!

tests/test_app.rs
```diff
@@ -610,13 +689,40 @@ fn test_impl_traits() {
...
+            let body = serde_json::to_string(&self).unwrap();
+
+            // Create response and set content type
+            ready(Ok(actix_web::HttpResponse::Ok()
+                .content_type("application/json")
+                .body(body)))
+        }
+    }
+
+    /// TODO: Returning impl Responder will not output any schema. How to tell what really function returns?
```
wafflespeanut: I don't think we can get around this problem as of now.

Issue #62, opened 2019-07-28
Actix plugin: Response codes other than 200


wafflespeanut:
With the current actix-web flow, we can (atmost) get two types of status codes through `Result<T, E>` and for everything else, we only have one status code. So, we need to think about how we can represent error models used by the server.

platy:
I just came across this limitation and am having a go at solving it for my use case.

There are more complicated use cases such as services which could return several different types of successful responses and in those cases it would be better to have real descriptions, but I don't think I'll need those any time soon.

The solution i'm attempting is a series of `struct`s implementing `Responder` and `Apiv2Schema` which repeat the implementations on `Json` + one for `StatusCode::NO_CONTENT`. I like that there is no ambiguity here, you either get the status code set on both the responder and on paperclip or on neither.

So a handler becomes:
```rust
#[api_v2_operation]
pub async fn handle() -> Result<CreatedJson<JustMessage>, Error> {
    Ok(CreatedJson(JustMessage {
        message: "Successful.",
    }))
}
```



Implementation in gist : https://gist.github.com/platy/e0bc44e0ce91c8268dcf0d9689e1f974

PR #158, opened 2020-03-02
Added api_v2_errors

sfisol:
I've added api_v2_errors macro which implements Apiv2Errors trait for selected struct - it can be then (afer implementing also ResponseError) used in handler return type (example added in tests) to document errors in swagger.

Additional implementation of OperationModifier on Result<T, E> where
T is OperationModifier and E is OprationErrorsModifier
allows to automatically add defined errocodes to swagger handler response.

I'll test it in my project if it will work smooth in following days, but meanwhile you could maybe take a look if this approach is good enough (it's not the best for sure :-))

### DIFF
--- a/book/actix-plugin.md
+++ b/book/actix-plugin.md
@@ -156,6 +156,55 @@ curl http://localhost:8080/api/spec
 
 Similarly, if we were to use other extractors like `web::Query<T>`, `web::Form<T>` or `web::Path`, the plugin will emit the corresponding specification as expected.
 
+#### Manually defining additional response codes
+
+There is a macro `api_v2_errors` which helps to manually add responses other than 200. Here is an example.
+We define an enum that represents the error. To use it in actix-web it needs to implement `Display` and `ResponseError`.
+There is no need to impl Display by hand for such simple enums - we can use `enum-display-derive` lib.
+
+```rust
+use actix_web::{
+    error::{ErrorInternalServerError, ErrorUnauthorized, ResponseError},
+    HttpResponse,
+};
+use enum_display_derive::Display;
+use paperclip::actix::api_v2_errors;
+use std::fmt::Display;
+
+#[api_v2_errors(
+    "400 Bad Request",
+    "401 Unauthorized: Can't read session from header",
+    "500 Internal Server Error",
+)]
+#[derive(Debug, Display)]
+pub enum UserError {
+    Unauth,
+    Internal(String),
+}
+
+impl ResponseError for UserError {
+    fn error_response(&self) -> HttpResponse {
+        match self {
+            Self::Unauth =>
+              HttpResponse::from_error(
+                  ErrorUnauthorized("Unauthorized: Can't read session from header")
+              ),
+            Self::Internal(msg) =>
+              HttpResponse::from_error(
+                  ErrorInternalServerError(msg.clone())
+              ),
+        }
+    }
+}
+```
+
+Such enum can then be used in handler return type to have defined response codes added to this handler docs:
+
+```rust
+#[api_v2_operation]
+async fn my_handler() -> Result<Pet, UserError> { Err(UserError::Internal("Some error")) }
+```
+
 #### Known limitations
 
 - **Enums:** OpenAPI (v2) itself supports using simple enums (i.e., with unit variants), but Rust and serde has support for variants with fields and tuples. I still haven't looked deep enough either to argue that this cannot be done in OpenAPI or find an elegant way to represent this in OpenAPI.
--- a/book/actix-plugin.md
+++ b/book/actix-plugin.md
@@ -158,51 +158,19 @@ Similarly, if we were to use other extractors like `web::Query<T>`, `web::Form<T
 
 #### Manually defining additional response codes
 
-There is a macro `api_v2_errors` which helps to manually add responses other than 200. Here is an example.
-We define an enum that represents the error. To use it in actix-web it needs to implement `Display` and `ResponseError`.
-There is no need to impl Display by hand for such simple enums - we can use `enum-display-derive` lib.
+There is a macro `api_v2_errors` which helps to manually add responses other than 200.
 
 ```rust
-use actix_web::{
-    error::{ErrorInternalServerError, ErrorUnauthorized, ResponseError},
-    HttpResponse,
-};
-use enum_display_derive::Display;
 use paperclip::actix::api_v2_errors;
-use std::fmt::Display;
 
 #[api_v2_errors(
-    "400 Bad Request",
-    "401 Unauthorized: Can't read session from header",
-    "500 Internal Server Error",
+    code=400,
+    code=401, description="Unauthorized: Can't read session from header",
+    code=500,
 )]
-#[derive(Debug, Display)]
-pub enum UserError {
-    Unauth,
-    Internal(String),
+pub enum MyError {
+    /* ... */
 }
-
-impl ResponseError for UserError {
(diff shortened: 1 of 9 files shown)


sfisol:
It's not very convenient to define struct and impl Display and ResponseError for it to just add errorcodes to swagger, but it worked for me well. There is a handy lib enum-display-derive which makes it more compact. I've added an example to the book.

wafflespeanut:
Thank you very much for your work! I'll look into this in detail tonight. I have a few suggestions based on how I'd planned for this.

book/actix-plugin.md
```diff
@@ -156,6 +156,55 @@ curl http://localhost:8080/api/spec
...
+    HttpResponse,
+};
+use enum_display_derive::Display;
+use paperclip::actix::api_v2_errors;
+use std::fmt::Display;
+
+#[api_v2_errors(
+    "400 Bad Request",
+    "401 Unauthorized: Can't read session from header",
+    "500 Internal Server Error",
```
wafflespeanut: I'm wary of using strings here. Firstly, can we make this a bit more readable?

```rust
#[api_v2_errors(
    code = 400, description = "Bad Request",
    code = 500, description = "Internal Server Error",
)]
```

I also don't want to enforce the description here. Users can avoid descriptions, in which case we'd fill the description of the status code automatically (from say, `http` crate).

```rust
#[api_v2_errors(
    code = 400, // description will default to "Bad Request"
    code = 500, description = "Booya",
)]
```

Users could also choose to not have any descriptions, in which case, they could go for a list of status codes.

```rust
#[api_v2_errors(400, 500)]
```

This is what I have in mind. If you have a suggestion, then I'm all ears! :smile:

core/src/v2/schema.rs
```diff
@@ -341,3 +341,11 @@ pub trait Apiv2Operation<T, R> {
     /// Returns the definitions used by this operation.
     fn definitions() -> BTreeMap<String, DefaultSchemaRaw>;
 }
+
+/// Represents a OpenAPI v2 error convertible. This is auto-implemented by
+/// [`api_v2_errors`](https://paperclip.waffles.space/paperclip_actix_macros/attr.api_v2_errors.html) macro.
+///
+/// This is implemented for primitive types by default.
+pub trait Apiv2Errors {
+    const ERROR_MAP: &'static [(&'static str, &'static str)];
```
wafflespeanut: This should be tuples of `u16` and `&'static str` instead.

macros/src/actix/mod.rs
```diff
@@ -90,6 +90,60 @@ pub fn emit_v2_operation(input: TokenStream) -> TokenStream {
...
+        .0
+        .iter()
+        .filter_map(|attr| match attr {
+            NestedMeta::Lit(Lit::Str(attr)) => {
+                let attr_value = attr.value();
+                let mut chunks = attr_value.splitn(2, ' ').map(str::to_string);
+
+                let error_code = chunks.next().unwrap_or_else(|| "".to_string());
+                let error_message = chunks.next().unwrap_or_else(|| "".to_string());
+
```
wafflespeanut: We have to ensure that the status code is valid (or throw error at compile time), given that we're should generate a valid OpenAPI spec.

book/actix-plugin.md
```diff
@@ -156,6 +156,55 @@ curl http://localhost:8080/api/spec
 
 Similarly, if we were to use other extractors like `web::Query<T>`, `web::Form<T>` or `web::Path`, the plugin will emit the corresponding specification as expected.
 
+#### Manually defining additional response codes
+
+There is a macro `api_v2_errors` which helps to manually add responses other than 200. Here is an example.
+We define an enum that represents the error. To use it in actix-web it needs to implement `Display` and `ResponseError`.
+There is no need to impl Display by hand for such simple enums - we can use `enum-display-derive` lib.
```
wafflespeanut: Let's just mention that we offer a macro which is useful for defining error status codes? I don't want to mention any specifics like implementing `ResponseError` on enums or using `enum-display-derive` here. We can assume that our users are already using actix and are just looking for automatic spec generation.

book/actix-plugin.md
```diff
@@ -156,6 +156,55 @@ curl http://localhost:8080/api/spec
...
+    "401 Unauthorized: Can't read session from header",
+    "500 Internal Server Error",
+)]
+#[derive(Debug, Display)]
+pub enum UserError {
+    Unauth,
+    Internal(String),
+}
+
+impl ResponseError for UserError {
```
wafflespeanut: Let's keep this subtle and remove `impl ResponseError for UserError` entirely.

sfisol: So do you think 'examples' directory could be a good place for this in future?

sfisol:
I don't think #62 can't be closed after this PR. You still won't be able to have 201 in swagger instead of 200 as 200 is inserted always.

sfisol:
```rust
#[api_v2_errors(
    description="foo",
    code=403, description="bar",
    code=-10,
    code="100", description="spam", description="eggs",
    code=50,
    code=469,
    code=1500,
    code=1000000000,
)]
```

Will generate bunch of errors and warnings:

```error: Attribute 'description' can be only placed after prior 'code' argument
  --> src/handlers/posts.rs:12:5
   |
12 |     description="adsfadsf",
   |     ^^^^^^^^^^^^^^^^^^^^^^

error: Invalid u16 in code argument
  --> src/handlers/posts.rs:14:5
   |
14 |     code=-10,
   |     ^^^^^^^^

error: Invalid macro attribute. Should be 'code = int' or 'description = str'
  --> src/handlers/posts.rs:15:5
   |
15 |     code="100", description="blablabla", description="datata",
   |     ^^^^^^^^^^

warning: This attribute overwrites previous description
  --> src/handlers/posts.rs:15:42
   |
15 |     code="100", description="blablabla", description="datata",
   |                                          ^^^^^^^^^^^^^^^^^^^^

error: Invalid u16 in code argument
  --> src/handlers/posts.rs:19:5
   |
19 |     code=1000000000,
   |     ^^^^^^^^^^^^^^^

warning: Invalid status code 50
  --> src/handlers/posts.rs:16:5
   |
16 |     code=50,
   |     ^^^^^^^

warning: Status code 469 doesn't have a canonical name
  --> src/handlers/posts.rs:17:5
   |
17 |     code=469,
   |     ^^^^^^^^

warning: Invalid status code 1500
  --> src/handlers/posts.rs:18:5
   |
18 |     code=1500,
   |     ^^^^^^^^^

error: aborting due to 4 previous errors
```

Let me know is you want maybe some of warnings be changed to errors, for invalid codes for example.

wafflespeanut:
> I don't think #62 can be closed after this PR.

You're right! Should probably just leave it as a link rather than link it through Github.

> Will generate bunch of errors and warnings:

So cool! Nice work! Thanks!

book/actix-plugin.md
wafflespeanut: Yeah, we'll take care of detailed examples in the future. Right now, this feels redundant.

wafflespeanut:
Thank you very much for working on this!

sfisol:
Boohoo, you've overwritten my authorship :stuck_out_tongue: 

Next, I'd try to add tags in operation macro attribute :-)

wafflespeanut:
I've no idea how it happened! I merged it the same way as your previous PR! But anyway, I've done a rebase, changed the commit author and force pushed! Your authorship has been restored!

Looking forward to your work! :smile:

PR #168, opened 2020-03-27
Update nightly version

oli-cosmian:


wafflespeanut:
We should also update `.travis.yml`

oli-cosmian:
done

wafflespeanut:
Thanks!

PR #169, opened 2020-03-27
Use a derive proc macro instead of an attribute proc macro

oli-cosmian:
The input was never modified, so it seemed better to make it a derive macro. The major change compared to master is is that you need to specify `#[openapi(empty)]` in addition to `#[derive(Apiv2Schema)]` if you want an `Any` type

### DIFF
--- a/macros/src/actix.rs
+++ b/macros/src/actix.rs
@@ -9,8 +9,9 @@ use quote::quote;
 use strum_macros::EnumString;
 use syn::spanned::Spanned;
 use syn::{
-    Attribute, Data, DataEnum, Field, Fields, FieldsNamed, FieldsUnnamed, Generics, Ident, ItemFn,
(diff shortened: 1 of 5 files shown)


wafflespeanut:
Makes sense, but can't we have something like this?

```rust
#[derive(Apiv2Schema)]
#[openapi(empty)]
struct Foo {
    field: f32,
}
```

This is similar to what `serde` does, for type-level attributes, and we need it, because skipping fields is one thing, but an empty schema in OpenAPI corresponds to an `Any` type. In our case, this isn't `Any`, but rather something that we don't include in the generated spec. We'd also need type-level attributes in the future for other things.

oli-cosmian:
I removed the skipping fields feature, I'll readd it in a separate PR. For now this PR does not change any behaviour, it's a 1:1 replacement for the existing attribute.

wafflespeanut:
Awesome! Thank you!

PR #172, opened 2020-03-30
Harden `paperclip` against other crates using the `kv_unstable` feature of `log`

oli-cosmian:
Without these changes, `paperclip` won't compile if any crate in the dep tree uses that feature. I made sure that the `kv_unstable` dependency is not part of `paperclip`, but only of the `dev-dependencies`, so we'll see the failure with `cargo test`, but not expose the feature to users of `paperclip`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ actix-web = "2.0"
 chrono = { version = "0.4", features = ["serde"] }
 futures = "0.3"
 uuid = { version = "0.8", features = ["serde"] }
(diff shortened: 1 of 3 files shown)


wafflespeanut:
Damn. This was brought on by @sfisol in #139. I never realized that this could be an actual problem.

Btw, could you rustfmt the changes, and I'll merge it?

oli-cosmian:
done

wafflespeanut:
Thanks!

PR #175, opened 2020-03-31
Forward user docs

oli-cosmian:
Based on #169

Causes any documentation on `#[derive(Apiv2Schema)]` types to get exposed as an OpenApi `description` field.

### DIFF
--- a/core/src/v2/schema.rs
+++ b/core/src/v2/schema.rs
@@ -194,6 +194,9 @@ pub trait Apiv2Schema {
     /// Name of this schema. This is the object's name.
     const NAME: Option<&'static str> = None;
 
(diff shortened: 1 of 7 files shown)


wafflespeanut:
This is great! Could you also fix the clippy issue and we'll merge this?

oli-cosmian:
The issue is that current nightly has no rustfmt. The only way to fix it permanently is to make nightly CI not run rustfmt and clippy I think.

wafflespeanut:
I didn't look into the error earlier. It's fine actually. I've switched to stable channel in Travis lately. Both rustfmt and clippy run in stable as of now. Your PR is failing because it's not in sync with master. :smile:

PR #176, opened 2020-03-31
Security definitions

sfisol:
- Added api_v2_security macro for actix plugin

### DIFF
--- a/book/actix-plugin.md
+++ b/book/actix-plugin.md
@@ -193,6 +193,53 @@ async fn my_handler() -> Result<(), MyError> {
 }
 ```
 
+#### Defining security
+
+Use `Apiv2Security` derive macro for struct used as handler parameter to have this handler marked as requiring authorization.
+
+```rust
+use paperclip::actix::Apiv2Security;
+
+#[derive(Apiv2Security)]
+#[openapi(
+  apiKey,
+  in = "header",
+  name = "Authorization",
+  description = "Use format 'Bearer TOKEN'"
+)]
+pub struct AccessToken;
+
+impl FromRequest for Accesstoken { /*...*/ }
+
+#[api_v2_operation]
+async fn my_handler(access_token: AccessToken) -> Result {
+    /*...*/
+}
+```
+
+First parameter is the type of security, currently supported types are "apiKey" and "oauth2". Possible parameters are `alias`, `description`, `name`, `in`, `flow`, `auth_url`, `token_url` or `parent`.
+
+Use `alias` parameter if you need to have two different security definitions of the same type.
+
+If you need to define scopes for `oauth2`, use `parent` attribute:
+
+```rust
+#[derive(Apiv2Security, Deserialize)]
+#[openapi(
+  oauth2,
+  auth_url = "http://example.com/",
+  token_url = "http://example.com/token",
+  flow = "password"
+)]
+struct OAuth2Access;
+
+#[derive(Apiv2Security, Deserialize)]
+#[openapi(parent = "oauth2", scopes("pets.read", "pets.write"))]
+struct PetScopeAccess;
+```
+
+Keep in mind that the parent struct must be used in a mounted handler so paperclip could generate proper security definition when generating spec.
+
 #### Known limitations
 
 - **Enums:** OpenAPI (v2) itself supports using simple enums (i.e., with unit variants), but Rust and serde has support for variants with fields and tuples. I still haven't looked deep enough either to say whether this can/cannot be done in OpenAPI or find an elegant way to represent this in OpenAPI.
--- a/core/src/v2/models.rs
+++ b/core/src/v2/models.rs
@@ -15,7 +15,7 @@ use regex::{Captures, Regex};
 use actix_http::http::Method;
 
 use std::borrow::Cow;
-use std::collections::{BTreeMap, BTreeSet};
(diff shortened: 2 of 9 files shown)


wafflespeanut:
Hi! Is there any leftover work to do? Or is this ready for review?

sfisol:
Hi!
Sorry, I haven't time to write anything here in PR. So, this is all I need currently in my work. I was thinking of adding something to docs, but am I now in a hurry... If you're happy with the changes then review and eventually merge. If not then I can add/change/doc something by the end of the week.
Cheers!

wafflespeanut:
Okay, thank you very much for this work! But, I'll need to think about how/whether this could be done in a better way. I'll review this in the weekend.

sfisol:
Hi, what's up? :-)
I could probably refactor something this weekend if you have any remarks.

wafflespeanut:
Hey, sorry for the late response. Here are my thoughts.

Firstly, I don't like that we're adding the security definitions manually. We can automatically add them when we walk through the operations. This is how we take care of schema definitions (i.e., serde structs). I think the usage should be more like:

```rust
// NOTE: "apiKey" and "header" correspond to "type" and "in" keys.
#[derive(Apiv2Security)]
#[openapi(apiKey, header = "X-My-Api-Key")]
struct ApiKeyAuth;  // Name of the struct becomes name of the definition
```

> Note that I've switched to using derive macro instead of attribute macro (since we'd be using this to mark structs and not functions).

Whenever we encounter a struct in a handler, we can update the global security definition. While this works for `basic` and `apiKey` auth, this won't work for OAuth2. Following what we have above, we can track `scopes`, but in order to generate a valid swagger spec, we also have to track the `flow`, `authorizationUrl` and `tokenUrl` fields. So, I propose the following.

```rust
#[derive(Apiv2Security)]
#[openapi(oAuth2, tokenUrl = "...", authUrl = "...", flow = "...")]
struct MyOAuth2;  // Name of struct becomes name of definition

#[derive(Apiv2Security)]
#[openapi(parent = MyOAuth2, scopes("pets.read", "pets.write"))]
struct PetScope;
```

core/src/v2/schema.rs
```diff
@@ -184,6 +184,8 @@ impl_type_simple!(
 #[cfg(feature = "uid")]
 impl_type_simple!(uuid::Uuid, DataType::String, DataTypeFormat::Uuid);
 
+pub type SchemaSecurityDefs = &'static [(&'static str, &'static [&'static str])];
```
wafflespeanut, line 187: I also don't like that we have this complex structure. Either way, all of this happens at runtime. So, we could just have an associated function to emit the information we need (just like `raw_schema` function).

wafflespeanut:
Also, thank you very much for your contributions. Feel free to let me know your thoughts on this. :smile:

sfisol:
I agree with your ideas!  :+1: 
I'll amend PR today or tomorrow.

sfisol:
@wafflespeanut Can you probably take a quick review of the changes now? It's still a draft, macro is cumbersome and unfinished, but you'll maybe have some remarks on the definition and use of `resolve_security_defs` function.

I'll be able to continue in a few days.

core/src/v2/models.rs
```diff
@@ -569,6 +586,10 @@ pub struct Operation<P, R> {
     pub consumes: Option<BTreeSet<MediaRange>>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub produces: Option<BTreeSet<MediaRange>>,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub security: Vec<BTreeMap<String, Vec<String>>>,
+    #[serde(skip)]
+    pub security_definitions: BTreeMap<String, SecurityScheme>,
```
wafflespeanut, line 592: I think the only reason this exists is  because one of the updating functions have this signature:
```rust
fn update_security(op: &mut DefaultOperationRaw) {}
```
I'll take care of cleaning this up later.

wafflespeanut: And you had to do this because the operation function itself has the following signature:
```rust
fn operation() -> DefaultOperationRaw;
```

wafflespeanut: Feel free to finish up with the implementation. I think it's better if I refactor everything and have a global context to take care of some stuff at the operation level.

book/actix-plugin.md
```diff
@@ -193,6 +193,53 @@ async fn my_handler() -> Result<(), MyError> {
...
+#[openapi(
+  oauth2,
+  auth_url = "http://example.com/",
+  token_url = "http://example.com/token",
+  flow = "password"
+)]
+struct OAuth2Access;
+
+#[derive(Apiv2Security, Deserialize)]
+#[openapi(parent = "oauth2", scopes("pets.read", "pets.write"))]
```
wafflespeanut, line 237: What is the motivation behind `parent = "oauth2"`? Can't we have `parent = OAuth2Access`? Because, it shouldn't be the case that `OAuth2Access` must be used in some handler. If we had the latter, then we can generate the definition if it doesn't already exist.

wafflespeanut:
Sorry I missed this for a long time. Feel free to ping me next time (here or in chat), just in case I've missed the email. :sweat_smile:

wafflespeanut:
Thanks for the PR! Great work! :smile:

PR #181, opened 2020-05-15
Added impl of DataTime<Utc> to Apiv2Schema

sfisol:


core/src/v2/actix.rs
```diff
@@ -155,6 +155,8 @@ mod manual_impl {
     impl_simple!(usize);
     #[cfg(feature = "datetime")]
     impl_simple!(chrono::NaiveDateTime);
+    #[cfg(feature = "datetime")]
+    impl_simple!(chrono::DateTime<chrono::Utc>);
```
wafflespeanut, line 159: Can we implement this for generic `DateTime<T>` rather than just for `DateTime<Utc>`?

wafflespeanut:
Closed and reopened so that Travis picks it up.

sfisol:
Done, but keep in mind that Swagger will render it wrongly as UTC (f.ex. "2020-05-26T10:14:40.337Z") just as it renders it wrongly for NaiveDateTime :-)

wafflespeanut:
Note that the rendering doesn't have anything to do with the plugin itself (it's the serde impl). We only implement a trait for emitting the schema.

PR #182, opened 2020-05-15
Impl OperationModifier for actix_multipart

sfisol:
I'm not sure if these types covers all possible multipart cases, but better this than nothing :-)

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -12,6 +12,7 @@ repository = "https://github.com/wafflespeanut/paperclip"
 [dependencies]
 actix-web = { version = "2.0", optional = true }
 actix-http = { version = "1.0", optional = true }
+actix-multipart = { version = "0.2", optional = true }
 chrono = { version = "0", optional = true }
 failure = "0.1"
 heck = { version = "0.3", optional = true }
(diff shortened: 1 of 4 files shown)


core/Cargo.toml
```diff
@@ -28,7 +29,7 @@ serde_yaml = "0.8"
 uuid = { version = "0", optional = true }
 
 [features]
-actix = ["v2", "actix-http", "actix-web"]
+actix = ["v2", "actix-http", "actix-web", "actix-multipart"]
```
wafflespeanut, line 32: Let's have a `multipart` feature instead, which depends on `actix-multipart` please? Like, how we have it for `chrono` and `datetime` (the dependency should be opt-in).

wafflespeanut:
Closed and reopened so that Travis picks it up.

core/Cargo.toml
```diff
@@ -12,6 +12,7 @@ repository = "https://github.com/wafflespeanut/paperclip"
 [dependencies]
 actix-web = { version = "2.0", optional = true }
 actix-http = { version = "1.0", optional = true }
+actix-multipart = { version = "0.2", optional = true }
```
wafflespeanut, line 15: Does this work at the moment? There's no feature in `core` that ties this dependency.

sfisol: It seems that if there is no feature then plain optional dependency is activated. I haven't found any info in cargo documentation about this, but I've checked it and it works for me.

wafflespeanut: Interesting. I guess this is why they don't allow feature name to collide with dependency name. :smile:

PR #191, opened 2020-05-31
Migrate from failure

MikailBag:
`failure` is deprecated, and `thiserror` + `anyhow` combination is the most similar migration target.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,6 @@ required-features = ["cli"]
 
 [dependencies]
 env_logger = { version = "0.7", optional = true }
-failure = "0.1"
-failure_derive = "0.1"
 git2 = { version = "0.13", optional = true }
 heck = { version = "0.3", optional = true }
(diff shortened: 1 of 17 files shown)


MikailBag:
I don't understand why tests/test_k8s/Cargo.toml is not updated (still requires failure), but tests/test_k8s/cli/Cargo.toml is OK. Do they both use `src/build/manifest.hbs`?

wafflespeanut:
No, the former is [in tree](https://github.com/wafflespeanut/paperclip/tree/e537a28bf161c730e5ab0e5aab460706377151d4/tests/test_k8s), because it's to test the build script way of codegen (which doesn't emit `lib.rs` and `Cargo.toml`). The latter is the CLI-way of codegen (where both `lib.rs` and `Cargo.toml` are emitted). :smile:

wafflespeanut:
This is amazing! Thank you very much for your work! :smile:

Issue #37, opened 2019-07-08
Support basic auth if specified in security definitions


wafflespeanut:
Currently, we support adding root certs and enabling client verification, but we also need basic auth. We should support that in the generated client if the field is specified and add an option to CLI regardless.

letmutx:
I would like to work on this.

wafflespeanut:
Cool!

wafflespeanut:
So, this is basically what I had in mind.

The two traits of interest are `ApiClient` and `Sendable`.

```rust
trait ApiClient {
    //
}
```

```rust
trait Sendable {

    type Output: DeserializeOwned + Send + 'static;
    
    // other thingies ...
    
    fn send(&self, client: &dyn ApiClient) -> Box<...> {
        //
    }
}
```

> A number of things here should be type/generic parameters, but they're coupled to `reqwest::async::*` as of now (which is fine, because `reqwest` doesn't have a competition).

`Sendable` is implemented for all "fulfilled" builder structs i.e., builders which have the necessary parameters for making an API call. So, each impl represents an operation.

In OpenAPI, auth can be imposed on all operations or one particular operation. In order to add type-level constraint for auth, we should extend `ApiClient` for security stuff.

```rust
trait ApiClient<Auth> {
    //
}

struct NoAuth;
```

By default, `reqwest::async::Client` (which currently impls `ApiClient`) will now impl `ApiClient<NoAuth>`. We'll have an extension trait for transforming the default client to other kinds of clients.

```rust
struct AuthBasic;
struct AuthApiKey;

trait ClientExt {
    type BasicAuthClient: ApiClient<AuthBasic>;
    type ApiKeyClient: ApiClient<AuthApiKey>;
    // Other things (like OAuth) will follow later.
    
    fn with_basic_auth(&self, user: &str, pass: &str) -> Self::BasicAuthClient;
    
    fn with_api_key(&self, key: &str) -> Self::ApiKeyClient;
}
```

Then, we'll have wrappers for the actual client which store auth data.

```rust
struct ReqwestBasicAuth {
    user: String,
    pass: String,
    client: reqwest::async::Client;
}

impl ApiClient<AuthBasic> for ReqwestBasicAuth {
    // `make_request` will call `make_request` on the inner client
    // and will set the basic auth on the emitted request object.
}
```

[...]

PR #193, opened 2020-06-02
Make custom client

MikailBag:
I have not tested these changes, but goal is to allow setting API base path at runtime.


MikailBag:
I also added basic api key support (it seems paperclip currently does not support authentication, does it?)

wafflespeanut:
Paperclip doesn't have auth support yet, but I do have a plan to work on this before 0.4.0 release (to enforce auth for certain operations at compile time detailed here - https://github.com/wafflespeanut/paperclip/issues/37#issuecomment-538771621).

PR #195, opened 2020-06-02
Emit helper unit types in reproducible order

MikailBag:


MikailBag:
Now CI passes

wafflespeanut:
Thank you for looking around and cleaning up! :smile:

PR #198, opened 2020-07-01
Fix a mis-namespaced TokenStream ref

mcsaucy:
This should fix the broken build.

I was trying to figure out how this ever worked, and why it suddenly broke. Looks like `proc-macro-error` introduced a "breaking" (IMO this never should have worked) change from 1.0.2 to 1.0.3, as everything is green if you pin `proc-macro-error` to `=1.0.2`. I'm an absolute novice here, but [this jumped out](https://gitlab.com/CreepySkeleton/proc-macro-error/-/commit/141ebd775d63ada98d9f17eca8762d6c3522d55e).

### DIFF
--- a/macros/src/actix.rs
+++ b/macros/src/actix.rs
@@ -397,7 +397,7 @@ pub fn emit_v2_security(input: TokenStream) -> TokenStream {
         }
     }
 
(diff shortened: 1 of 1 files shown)


frol:
@wafflespeanut can you make a patch release for this?

wafflespeanut:
Yes! I'm planning to do it tonight!

PR #211, opened 2020-08-15
Add operation descriptions and respect generics via api_v2_operation macros

dunnock:
This PR adds operation descriptions (e.g. summary and description fields of Operation) generated from doc comments by api_v2_operation macros.
It also fixes usage of generics in handler definition, so that unit struct would inherit them. The only inconvenience with generics is they should implement Apiv2Schema and in the operation description one should provide bounds:
```
    #[api_v2_operation]
    async fn get_pet_by_name<S: paperclip::v2::schema::Apiv2Schema + ToString>(
        _path: web::Path<S>,
    ) -> Result<web::Json<Vec<Pet>>, ()> {
        Ok(web::Json(vec![Pet::default()]))
    }
```

Added relevant test cases respectively.


wafflespeanut:
Thank you very much for initiating this. I have an idea for refactoring the plugin code, so I'll hold this off for the time being. I'll have an answer in the next 2 days. :smile:

wafflespeanut:
Thank you for coming up with this PoC and starting this discussion. But, I'd like to refactor the existing logic for deriving `Apiv2Operation` (which is getting a bit of a technical debt) before introducing more things that rely on it.

As of now, there's an `OperationModifier` which is implemented (internally) for all handler-like functions. The problem with the existing approach is that functions with the same signatures will generate the same spec (regardless of their names), which makes it impossible to tie operation-specific fields like summary, description and other overrides (that may appear in the future) directly, and we have to come up with more hackery. I think we could do this in a better way.

Currently, we have a `ResponseWrapper` type which is only used for `impl Responder` stuff, but we can change that to:
```rust
pub struct ResponseWrapper<T: Responder, H>(T, PhantomData<H>);
```

Now, let's take this simple handler:
```rust
#[api_v2_operation]
fn my_handler() -> Foo {
    //
}
```
This could generate the following code:
```rust
struct paperclip__my_handler;  // marker struct

fn my_handler() -> ResponseWrapper<paperclip__my_handler, Foo> {
    //
}
```
> This also opens the possibility to implement `Apiv2Operation` on the fly, instead of implementing it internally for functions.

Since we're using marker structs, we can now tie operation-specific metadata directly to the response wrappers (with the marker struct associated with the handler), and then we can use them in resources to get the operation data.

Once I refactor the `OperationModifier`, it'll be easier to implement the summary and description (without having to box stuff), at which point I'd be very happy to land this PR with the appropriate changes. What do you think?

dunnock:
Thank you, I see your point regarding ResponseWrapper, yeah that would be better fit in current architecture. Interesting ref refactoring of OperationModifier.

I will take amending this PR in a no rush mode. For us no rush atm as this workaround is working for us from a branch build and you anyways going to refactor OperationModifier first before merging this one.

wafflespeanut:
I've refactored the operation-deriving logic. It should now be easier to implement operation-specific schema changes. I've also left out summary and description, so you just have to rebase your PR and make some tiny changes. :smile:

dunnock:
@wafflespeanut I have cleaned up old code and implemented according to the new architecture. Apart of adding description and summary this PR also fixes usage of generics in handler definition. Please review

dunnock:
Fixed clippy

macros/src/actix.rs
```diff
@@ -129,23 +149,37 @@ pub fn emit_v2_operation(attrs: TokenStream, input: TokenStream) -> TokenStream
...
 
-    // panic!("{}",
+    let docs = extract_documentation(&item_ast.attrs);
+    let mut lines = docs.lines();
+    let summary = lines
+        .next()
+        .map(|line| quote!(Some(#line.to_string())))
+        .unwrap_or(quote!(None));
+    let description = lines.collect::<String>().trim().to_string();
+    let description = if !description.is_empty() {
```
wafflespeanut, line 166: Let's also expect a newline between summary and description for the time being? (as it's consistent with Rust docs). Also, this will be our last resort eventually, because I really want to make use of the attribute macro to achieve this (like `#[api_v2_operation(summary = "...", description = "...")]`

macros/src/actix.rs
```diff
@@ -42,6 +43,26 @@ pub fn emit_v2_operation(attrs: TokenStream, input: TokenStream) -> TokenStream
...
+                    Some(Pair::new(gen.ident.clone(), Some(*punct)))
+                }
+                Pair::End(syn::GenericParam::Type(gen)) => Some(Pair::new(gen.ident.clone(), None)),
+                _ => None,
+            })
+            .collect();
+        generics_call = quote!(::<#params> { p: std::marker::PhantomData });
+        struct_generics = quote!(<#params>);
+        struct_definition =
+            quote!(struct #unit_struct <#params> { p: std::marker::PhantomData<(#params)> } )
```
wafflespeanut, line 64: This is a nice catch!

macros/src/actix.rs
dunnock: Neat point, pushed a fix

wafflespeanut (approved):
Thank you for working on this!

PR #212, opened 2020-08-26
Attempts to implement serde_qs support

jdeepee:
Currently does not generate any query arguments where used in my application. But it does allow me to compile where serde_qs is used. Perhaps this can be the start to a proper implementation of serde_qs. 

Any guidance on what is required to get it correctly generating query arguments in the resulting API spec would be great. 

Thanks!


core/Cargo.toml
```diff
@@ -28,9 +28,11 @@ serde_json = "1.0"
...
 uuid = { version = "0", optional = true }
 thiserror = "1.0.19"
+serde_qs = { version = "0.6", optional = true, features = ["actix"] }
 
 [features]
 actix = ["v2", "actix-http", "actix-web", "pin-project"]
 codegen = ["v2", "heck", "log"]
 nightly = ["paperclip-macros/nightly"]
 v2 = ["paperclip-macros/v2"]
+qs = ["serde_qs"]
```
wafflespeanut, line 38: Do we need this? We can directly use dependency names for features. Besides, you're doing exactly that below.

plugins/actix-web/Cargo.toml
```diff
@@ -18,6 +18,8 @@ paperclip-macros = { path = "../../macros", version = "0.4.0", features = ["acti
 parking_lot = ">=0.10,<0.12"
 serde_json = "1.0"
 once_cell = "1.4.0"
+serde_qs = { version = "0.6", optional = true, features = ["actix"] }
 
 [features]
 nightly = ["paperclip-core/actix", "paperclip-core/nightly"]
+qs = ["serde_qs"]
```
wafflespeanut, line 25: newline please

plugins/actix-web/src/web.rs
```diff
@@ -4,6 +4,8 @@ pub use actix_web::web::{
     block, service, to, Bytes, BytesMut, Data, Form, FormConfig, HttpRequest, HttpResponse, Json,
     JsonConfig, Path, PathConfig, Payload, PayloadConfig, Query, QueryConfig,
 };
+#[cfg(feature = "qs")]
+pub use serde_qs::actix::QsQuery;
```
wafflespeanut, line 8: Let's not do this for now.

core/src/v2/models.rs
```diff
@@ -572,6 +572,7 @@ where
 #[serde(rename_all = "camelCase")]
 pub enum ParameterIn {
     Query,
+    QsQuery,
```
wafflespeanut, line 575: Please remove this. This is not conformant to the spec.

core/src/v2/actix.rs
```diff
@@ -337,6 +340,8 @@ impl<T: Apiv2Schema> Apiv2Schema for Form<T> {
 impl_param_extractor!(Path<T> => Path);
 impl_param_extractor!(Query<T> => Query);
 impl_param_extractor!(Form<T> => FormData);
+#[cfg(feature = "serde_qs")]
+impl_param_extractor!(QsQuery<T> => QsQuery);
```
wafflespeanut, line 344: This should be `impl_param_extractor!(QsQuery<T> => Query);`

wafflespeanut (changes requested):

core/Cargo.toml
jdeepee: Ahh didn't notice you could do that, thanks, updated

wafflespeanut (approved):
Thank you!

PR #213, opened 2020-09-02
Fix shared parameters generation

dunnock:
This PR fixes issue and one nit related to shared parameters generation:

- body is usually unique per method, hence do not generalize on body (e.g. patch has all fields optional)

- do not generalize when path has single method. also this generalized parameters feature nevertheless is part of a spec having some issues with some renderers, like redoc.

### DIFF
--- a/core/src/v2/models.rs
+++ b/core/src/v2/models.rs
@@ -346,12 +346,21 @@ impl<S> PathItem<Parameter<S>, Response<S>> {
         // We're using `Option<BTreeSet>` over `BTreeSet` because we need to
         // differentiate between the first operation that we use for initial
         // value of the set and  an operation that doesn't have any parameters.
+        if self.methods.len() < 2 {
+            return;
+        }
         let mut shared_params = None;
         for op in self.methods.values() {
             let params = op
                 .parameters
                 .iter()
-                .map(|p| p.name.clone())
+                .filter_map(|p| {
+                    if p.name != "body" {
+                        Some(p.name.clone())
+                    } else {
+                        None
+                    }
+                })
                 .collect::<BTreeSet<_>>();
             if let Some(p) = shared_params.take() {
                 shared_params = Some(&p & &params); // set intersection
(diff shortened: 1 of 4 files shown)


wafflespeanut:
> body is usually unique per method, hence do not generalize on body (e.g. patch has all fields optional)

We don't generalize unless we find the same body in all operation objects in that route. This is allowed by the spec.

> do not generalize when path has single method. also this generalized parameters feature nevertheless is part of a spec having some issues with some renderers, like redoc.

Does it make sense to disable this normalization through a config parameter? We don't support configuration for framework plugins as of now, but we can.

dunnock:
> We don't generalize unless we find the same body in all operation objects in that route. This is allowed by the spec.

Oh I see, I did not knew about that there supposed to be such logic. Though I came from an issue which was reproducible in our project, so to illustrate I have commented out my change and amended params test. It has expected additional output:
```json
                  "/api/v2/{resource}/v": {
                    "parameters": [{
                      "format": "int32",
                      "in": "path",
                      "name": "resource",
                      "required": true,
                      "type": "integer"
                    }],
                    "post": {
                      "parameters": [{
                        "in": "body",
                        "name": "body",
                        "required": true,
                        "schema": {
                          "$ref": "#/definitions/BadgeBody"
                        }
                      }],
                      "responses": {}
                    },
                    "patch": {
                        "parameters": [{
                          "in": "body",
                          "name": "body",
                          "required": true,
                          "schema": {
                            "$ref": "#/definitions/BadgeBodyPatch"
                          }
                        }],
                        "responses": {}
                    }
                  },
```
Instead it produces:
```json
       "/api/v2/{resource}/v":{
         "parameters":[
            {
               "in":"body",
               "name":"body",
               "required":true,
               "schema":{
                  "$ref":"#/definitions/BadgeBody"
               }
            },
            {
               "format":"int32",
               "in":"path",
               "name":"resource",
               "required":true,
               "type":"integer"
            }
         ],
         "patch":{
            "responses":{
               
            }
         },
         "post":{
            "responses":{
               
            }
         }
      },
```

> Does it make sense to disable this normalization through a config parameter? We don't support configuration for framework plugins as of now, but we can.

That would be awesome, definitely makes sense.

wafflespeanut:
For the time being, we can disable normalization entirely, and we can introduce a mechanism (in the future) which performs a "deep equal" before marking the parameters as *shared*.

dunnock:
@wafflespeanut I have reimplemented it the way you described, just hidden normalization behind `normalize` feature (which is not usable atm), also amended tests accordnigly

wafflespeanut (approved):
Thanks for working on this!

Issue #216, opened 2020-09-15
actix: error[E0373]: closure may outlive borrowed value when using #[api_v2_operation]


strohel:
paperclip version: 90038bfd (current `master`, to be v0.5)

user code:
```rust
/// The `/city/v1/get` endpoint. HTTP request: [`CityQuery`], response: [`CityResponse`].
///
/// Get city of given ID localized to given language.
#[api_v2_operation]
pub(crate) async fn get(query: Query<CityQuery>, app: Data<AppState>) -> JsonResult<CityResponse> {
    let locations_es_repo = LocationsElasticRepository(app.get_ref());
    let es_city = locations_es_repo.get_city(query.id).await?;

    Ok(Json(es_city.into_resp(app.get_ref(), query.language).await?))
}
```

This seems to work fine with paperclip v0.4.1, but if I try to update to current master, I get the following lifetime error:
```
error[E0373]: closure may outlive the current function, but it borrows `app`, which is owned by the current function
  --> src/handlers/city.rs:47:1
   |
47 | #[api_v2_operation]
   | ^^^^^^^^^^^^^^^^^^^
   | |
   | may outlive borrowed value `app`
   | in this macro invocation
48 | pub(crate) async fn get(query: Query<CityQuery>, app: Data<AppState>) -> JsonResult<CityResponse> {
49 |     let locations_es_repo = LocationsElasticRepository(app.get_ref());
   |                                                        --- `app` is borrowed here
   | 
  ::: /home/strohel/.cargo/git/checkouts/paperclip-bb3ea4fb155f1f50/90038bf/macros/src/lib.rs:39:1
   |
39 | #[proc_macro_error]
   | ------------------- in this expansion of `#[api_v2_operation]`
   |
note: closure is returned here
  --> src/handlers/city.rs:47:1
   |
47 | #[api_v2_operation]
   | ^^^^^^^^^^^^^^^^^^^ in this macro invocation
   | 
  ::: /home/strohel/.cargo/git/checkouts/paperclip-bb3ea4fb155f1f50/90038bf/macros/src/lib.rs:39:1
   |
39 | #[proc_macro_error]
   | ------------------- in this expansion of `#[api_v2_operation]`
help: to force the closure to take ownership of `app` (and any other referenced variables), use the `move` keyword
   |
47 | move #[api_v2_operation]
   |
```

User code with expanded macros:
```rust
        struct paperclip_get;
        #[doc =
[...]

PR #217, opened 2020-09-16
Initialize Operation parameters via macro attributes

dunnock:
Some Operation parameters might be initialized via macro attributes, I believe this was discussed prior.

This change adds parsing of api_v2_operation macro attributes, it will validate if attribute can be initialized via macro parameter it will add that initialization to the operation init code.

Supported set of attributes:
- summary
- description
- operation_id
- consumes
- produces

If summary attribute is empty it will try to obtain summary and description from a function doc comments.

### DIFF
--- a/macros/src/actix.rs
+++ b/macros/src/actix.rs
@@ -116,7 +116,7 @@ pub fn emit_v2_operation(attrs: TokenStream, input: TokenStream) -> TokenStream 
     } else if is_impl_trait {
         quote!((|| #block)())
     } else {
-        quote!((|| async #block)())
+        quote!(async move #block)
     };
 
     item_ast.block = Box::new(
--- a/macros/src/actix.rs
+++ b/macros/src/actix.rs
@@ -227,6 +227,8 @@ fn extract_fn_arguments_types(item_ast: &ItemFn) -> Vec<Type> {
 
 /// Parse macro attrs, matching to Operation fields
 /// Returning operation attribute identifier and value initialization arrays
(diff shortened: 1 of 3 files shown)


dunnock:
I can also quickly add move to fix that bug #216 here unless you do it separately?

wafflespeanut:
Feel free to!

dunnock:
Done. I found that this issue may apply only to async case, as in non async case we cannot move ownership into future

macros/src/actix.rs
```diff
@@ -140,7 +116,7 @@ pub fn emit_v2_operation(attrs: TokenStream, input: TokenStream) -> TokenStream
     } else if is_impl_trait {
         quote!((|| #block)())
     } else {
-        quote!((|| async #block)())
+        quote!(async move #block)
```
wafflespeanut, line 119: Can we make this a move closure? `move || #block` and `move || async #block`?

wafflespeanut (commented):
Looks okay overall. Just one comment.

macros/src/actix.rs
dunnock: Applied. Out of curiosity, what's the difference of using closure in this case?

wafflespeanut: We're storing the result in a wrapper (below). If we only have a block, then we can't support handlers which have early returns.

wafflespeanut (approved):
Splendid! Thank you very much for working on this :smile:

PR #218, opened 2020-09-16
Deps: allow also actix-web 3.0 in addition to 2.0 (for v0.4.*)

strohel:
Actix 3.0 has been recently released. It seems that paperclip (v0.4.1)
builds just fine against both actix-web 2.0 and 3.0
(and their dependent libs).

So just relax the dependency and let downstream crates choose which major
version of actix they pull.

I've tested this with downstream crates requiring both actix 2.0 and crates requiring 3.0,
works wel. YMMV.

This change is based (in git sense) on v0.4.1 tag rather than current master. Current master seems to contain feature changes to be released as v0.5. It would be cool if actix 3.0 support can be released earlier as v0.4.2 (v0.2.2). If you'd create a branch out of v0.4.1 (or a some later commit still slated for "stable" 0.4 branch), I'll change target of this PR and it should apply cleanly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ once_cell = "1.4.0"
 [dev-dependencies]
 actix-rt = "1.0"
 actix-service = "1.0"
-actix-web = "2.0"
+actix-web = ">=2.0,<4"
 chrono_dev = { version = "0.4", features = ["serde"], package = "chrono" }
 futures = "0.3"
 uuid_dev = { version = "0.8", features = ["serde"], package = "uuid" }
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -10,8 +10,8 @@ homepage = "https://github.com/wafflespeanut/paperclip"
 repository = "https://github.com/wafflespeanut/paperclip"
 
 [dependencies]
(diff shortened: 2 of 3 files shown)


strohel:
This would obsolete #214, #215.

wafflespeanut:
This is great news! I think we can release 0.4.2 with the existing changes after merging this PR. Could you rebase it?

strohel:
> I think we can release 0.4.2 with the existing changes after merging this PR. Could you rebase it?

Okay, I've rebased this on master. Note that current master contains (at least) one regression #216, and that versions are already bumped (0.4 -> 0.5, 0.3 -> 0.4, 0.2 -> 0.3).

strohel:
> Note that current master contains (at least) one regression #216 

...but I see this seems this already has a fix in @dunnock's repo: https://github.com/dunnock/paperclip/commit/178bf906 (that is part of 2 PR's here)

wafflespeanut:
I can change the versions back and do a minor bump instead. It was done only for convenience. There hasn't been any breaking changes so far.

strohel:
Thanks!

wafflespeanut:
Thank you! :smile:

PR #219, opened 2020-09-16
Actix operation macros to support tags and info

dunnock:
Built on top of #217 this change allows to configure list of tags, info and other Api specification defaults when using actix plugin. I believe this closes the loop to be able to build a nice looking spec, e.g. allowing to render APIs grouped by categories:

Screenshot 2020-09-16 at 18 54 40

To configure default I initially thought to add methods to App, but it is safer to configure defaults right when Api wrapper is created. Hence I've extended trait with additional method. This should provide full back compatibility.:

```
            let mut spec = DefaultApiRaw::default();
            spec.tags = vec![
              Tag {
                  name: "Dogs".to_string(),
                  description: Some("Images of dogs".to_string()),
                  external_docs: None,
              },
              Tag {
                  name: "Cats".to_string(),
                  description: Some("Images of cats".to_string()),
                  external_docs: None,
              },
              Tag {
                  name: "Cars".to_string(),
                  description: Some("Images of nice cars".to_string()),
                  external_docs: None,
              },
          ];
          spec.info = Info {
            version: "0.1".into(), 
            title: "Image server".into(), 
            ..Default::default()
          };

            App::new()
                .wrap_api_with_spec(spec)
                .with_json_spec_at("/api/spec")
                .service(web::resource("/images/pets").route(web::get().to(some_pets_images)))
                .build()
```

Tags supported by Operation and api_v2_operation macros:
```
    #[api_v2_operation(tags(Cats,Dogs))]
    fn some_pets_images() -> impl Future<Output = web::Json<Vec<Image>>> {
        ready(web::Json(Vec::new()))
    }
```

### DIFF
--- a/tests/test_app.rs
+++ b/tests/test_app.rs
@@ -1371,6 +1371,143 @@ fn test_operations_documentation() {
     );
 }
 
+#[test]
+#[allow(unreachable_code)]
+fn test_operations_macro_attributes() {
+    /// Index operation
+    ///
+    /// This doc comment will be overriden by macro attrs
+    #[api_v2_operation(
+        summary = "Root",
+        description = "Provides an empty value in response",
+        operation_id = "getIndex",
+        consumes = "application/json,text/plain",
+        produces = "text/plain"
+    )]
+    fn index() -> impl Responder {
+        ""
+    }
+
+    #[derive(Serialize, Deserialize, Apiv2Schema)]
+    struct Params {
+        limit: Option<u16>,
+    }
+
+    /// List all pets (in summary)
+    ///
+    /// This doc comment will be used in description
+    #[api_v2_operation(operation_id = "getPets")]
+    fn get_pets(
+        _data: web::Data<String>,
+        _q: web::Query<Params>,
+    ) -> impl Future<Output = Result<web::Json<Vec<Pet>>, ()>> {
+        if true {
+            // test for return in wrapper blocks (#75)
+            return futures::future::err(());
+        }
+
+        futures::future::err(())
+    }
+
+    run_and_check_app(
+        || {
+            App::new()
+                .wrap_api()
+                .with_json_spec_at("/api/spec")
+                .service(web::resource("/").route(web::get().to(index)))
+                .service(web::resource("/pets").route(web::get().to(get_pets)))
+                .build()
+        },
+        |addr| {
+            let resp = CLIENT
+                .get(&format!("http://{}/api/spec", addr))
+                .send()
+                .expect("request failed?");
+
+            check_json(
+                resp,
+                json!({
+                    "definitions": {
+                        "Pet": {
+                            "description": "Pets are awesome!",
+                            "properties": {
+                                "class": {
+                                "enum": ["dog", "cat", "other"],
+                                    "type":"string"
+                                },
+                                "id": {
+                                    "format": "int64",
+                                    "type": "integer"
+                                },
+                                "name": {
+                                    "description": "Pick a good one.",
+                                    "type": "string"
+                                },
+                                "updatedOn": {
+                                    "format": "date-time",
+                                    "type": "string"
+                                },
+                                "uuid":{
+                                    "format": "uuid",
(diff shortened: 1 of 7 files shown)


wafflespeanut:
Hey, is there anything blocking this? Is it ready for review?

dunnock:
> Hey, is there anything blocking this? Is it ready for review?

Hi, yeah, it's ready for review. Sorry was quite busy this week and did not toggle that.

dunnock:
Ah I see, I need to resolve few conflicts

wafflespeanut:
No worries! Take your time. I was just wondering :sweat_smile:

dunnock:
@wafflespeanut this PR is ready for review

tests/test_app.rs
```diff
@@ -1085,6 +1086,123 @@ fn test_list_in_out() {
...
 
+#[test]
+fn test_tags() {
+    #[derive(Serialize, Apiv2Schema)]
+    struct Image {
+        data: String,
+        id: u64,
+    }
+
+    #[api_v2_operation(tags = "cats,dogs")]
```
wafflespeanut, line 1097: I feel like we can add `#[api_v2_tag]` to make our lives easier (and it'll be similar to what we have for `Apiv2Security` derive).
```rust
#[api_v2_tag(name = "cat", description="foo", external_docs="bar")]  // optional name, description and external docs
struct Cat;  // By default, it'll take the struct's exact name
```
... and then we use it in operations like so,
```rust
#[api_v2_operation(tags(Cat, Dog, ...))]
fn my_handler() { /* */ }
```
What do you think? If you have other things to do, we can land this as it is and I'll work on it :smile:

dunnock: Yeah I like this deriving approach as it is more type safe. Though not fully understand how tags would end up in schema. Do you mean that each handler would return list of tags via some additional trait method like operation_tags() and the API would traverse all handlers to compose the full list.

I'm not sure when I will be able to dedicate time for this as we just passed alpha and maintenance is taking some extra time. But as soon as I got this will be on my list for sure.

But still considering above - before merging this in even as intermediate, it would be worth changing macros attributes to the ones which you suggested:
```
#[api_v2_operation(tags(Cat, Dog, ...))]
```
This should be quite reasonable effort which I will try to complete this week. Will just treat them as string for now in this PR but it will make them compatible with the structural tags as you suggested.

wafflespeanut: > Do you mean that each handler would return list of tags via some additional trait method like operation_tags() and the API would traverse all handlers to compose the full list.

That would be a way, yes. We already do this for security definitions and per-handler security stuff.

> This should be quite reasonable effort which I will try to complete this week.

Cool. I'm totally okay with landing this as an intermediate version and if you don't find time, I can implement this at some point soon. Let me know either ways.

platy: @dunnock if you don't have time for it I probably have time to make this change this week

dunnock: Thanks @platy ! I'll take it this week, though if you need it sooner pls feel free to PR in my branch.

wafflespeanut:
Hey @dunnock I couldn't find a way to message you personally, so I'll just put it here. I see you've been contributing to paperclip for some time now, so I've invited you as a collaborator. It's just additional privileges for you to review and merge stuff (if and whenever you have time and on the off-chance I take long time to respond). This is completely voluntary and is in no way an obligation. Thank you for all your work!

dunnock:
hi @wafflespeanut thank you for your trust, I have accepted that. It seems paperclip has got into spotlight recently. On this ticket - it was a busy week for me, I will finish this ticket today.

dunnock:
BTW, this PR would allow to generate "schemes" on the Api level, might help with https://github.com/wafflespeanut/paperclip/issues/248

tests/test_app.rs
dunnock: Pushed relevant changes

wafflespeanut (approved):

Issue #222, opened 2020-09-21
I couldn't make the actix example in the book work


omid:
This code in the book returns some errors!

https://github.com/wafflespeanut/paperclip/blob/d08e85d4930269befeb81f3bc08149936e915ee6/book/actix-plugin.md

```
use actix_web::{App, HttpServer};
use paperclip::actix::{
    // extension trait for actix_web::App and proc-macro attributes
    OpenApiExt, Apiv2Schema, api_v2_operation,
    // use this instead of actix_web::web
    web::{self, Json},
};
use serde::{Serialize, Deserialize};

// Mark containers (body, query, parameter, etc.) like so...
#[derive(Serialize, Deserialize, Apiv2Schema)]
struct Pet {
    name: String,
    id: Option<i64>,
}

// Mark operations like so...
#[api_v2_operation]
async fn echo_pet(body: Json<Pet>) -> Result<Json<Pet>, ()> {
    Ok(body)
}

#[actix_web::main]
async fn main() -> std::io::Result<()> {
    HttpServer::new(|| App::new()
        // Record services and routes from this line.
        .wrap_api()
        // Add routes like you normally do...
        .service(
            web::resource("/pets")
                .route(web::post().to(echo_pet))
        )
        // Mount the JSON spec at this path.
        .with_json_spec_at("/api/spec")

        // ... or if you wish to build the spec by yourself...

        // .with_raw_json_spec(|app, spec| {
        //     app.route("/api/spec", web::get().to(move || {
        //         actix_web::HttpResponse::Ok().json(&spec)
        //     }))
        // })

        // IMPORTANT: Build the app!
        .build()
    ).bind("127.0.0.1:8080")?
    .run().await
}
```

I'm using the feature "actix", even tested with what we have currently in master.
And I also replaced `api_v2_schema` with `Apiv2Schema`.

The error I can see is:
```
error[E0599]: no method named `wrap_api` found for struct `actix_web::app::App<actix_web::app_service::AppEntry, actix_http::body::Body>` in the current scope
  --> src/main.rs:25:10
   |
25 |         .wrap_api()
   |          ^^^^^^^^ method not found in `actix_web::app::App<actix_web::app_service::AppEntry, actix_http::body::Body>`
```

Anyway, this example is old.

wafflespeanut:
Hi. Which version of actix-web do you have in `Cargo.toml`?

omid:
I have actix-web="3.0"

[...]

PR #223, opened 2020-09-21
Support version 0.4 and actix-3

omid:


omid:
Will fix #222

book/actix-plugin.md
```diff
@@ -8,10 +8,10 @@ Let's start with a simple actix-web application. It has `actix-web` and `serde`
 # [package] ignored for brevity
 
 [dependencies]
-actix-web = "2.0"
+actix-web = "3.0"
```
wafflespeanut, line 11: Can you leave a comment here mentioning that 2.0 is also supported?

book/actix-plugin.md
```diff
@@ -30,10 +30,15 @@ struct Pet {
 }
 
 async fn echo_pet(body: Json<Pet>) -> Result<Json<Pet>, ()> {
-    Ok(body)
+    let new_pet = Pet {
+        name: format!("The New {}", body.name),
+        id: body.id
+    };
```
wafflespeanut, line 36: This really feels like an unnecessary change when we're trying to demo the plugin.

wafflespeanut (commented):
Two nits.

book/actix-plugin.md
omid: I wanted to show it that the response should be a JSON for sure! (Not a HttpResponse or ...) 
And how to create it!

Or I'm wrong.

book/actix-plugin.md
omid: done, can you check plz

book/actix-plugin.md
wafflespeanut: No, it's just that it's an echo handler, so why not respond with the body as it is?

book/actix-plugin.md
wafflespeanut: We can just leave it at, `actix-web 2.0 is also supported`

book/actix-plugin.md
omid: OK, done

book/actix-plugin.md
omid: OK, done

wafflespeanut (approved):
Thank you! :smile:

dunnock:
Hi, by some reason after this fix I can't make paperclip to pick actix-web 2.0, event after forcing that explicitly in my repo:
```
actix-web = { version = "2.0.0", default_features = false, optional = true }
```

I got in the cargo tree:
```
db v0.1.0 (/root/project/db)
├── actix-web v2.0.0
│   ├── actix-codec v0.2.0 (*)
│   ├── actix-http v1.0.1 (*)
│   ├── actix-macros v0.1.2 (*)
│   ├── actix-router v0.2.5
...
├── paperclip v0.4.2 (https://github.com/tari-labs/paperclip?branch=taris-v2#3ee6ecfc)
│   ├── anyhow v1.0.32
│   ├── itertools v0.9.0
│   │   └── either v1.6.1
│   ├── once_cell v1.4.1
│   ├── paperclip-actix v0.2.2 (https://github.com/tari-labs/paperclip?branch=taris-v2#3ee6ecfc)
│   │   ├── actix-service v1.0.6 (*)
│   │   ├── actix-web v3.0.2
```

Can you pls advice on how to force it to actix-web 2.0?

wafflespeanut:
It's not this commit, because it doesn't affect the code. It could be a regression from #218. Either way, I can reproduce this locally. Cargo probably couldn't reuse actix-web, maybe because of the diff of dependencies between actix-web 2 and 3. There are two ways to approach this. Either we find that dependency which is causing the problem and fix it in the tree (I can't guarantee about this though) or we can expose features like `actix2` and `actix3` with `actix` defaulting to the latest version.

dunnock:
Weird, I have amended version in a fork to depend on actix2 and it works nice. I guess if you would used some feature which is unique to actix3 that would not worked.

As to solution - I resolved that in our fork, and I guess most of people will upgrade to actix3. But I guess correct approach would be to keep 0.4 with actix2 and release 0.5 with actix3..?

dunnock:
Yeah and I did not mention how did I noticed that - actually we use `nightly-06-30` toolchain and actix3 brings time-0.2 which does not compile on that nightly, so actix-web3 might be considered as a breaking change for us https://github.com/time-rs/time/issues/281

wafflespeanut:
> I guess if you would used some feature which is unique to actix3 that would not worked.

But we haven't... We've disabled the default features.

> I guess correct approach would be to keep 0.4 with actix2 and release 0.5 with actix3..?

This will be a maintenance problem for us, because we'd still have to support actix-web 2.0 in the long run (at least, as long as it's maintained by actix community), which means any bug fixes will require a release in `0.4.x` in parallel with other releases. If we had features instead, we don't have to maintain multiple versions of paperclip and we can rely on everyone to switch to the latest version eventually.

dunnock:
I see your point ref maintenance, was not aware of necessity of support for actix2, yeah features definitely make sense now. Though I guess since it is breaking change, considering actix3 would be the default that should be paperclip 0.5? Since least that would happen to people using actix2 if they autoupgrade is that their API definitions will be compiled for actix3 not compatible with actix2, but in our case it has broke build.

wafflespeanut:
@dunnock I've now bumped minor version instead of patch version. Folks who use actix-web 2.0 can use `actix2` feature. The changes are in master now. Can you check?

PR #224, opened 2020-09-23
Accept both &str and String in the resource function of actix plugin

omid:


plugins/actix-web/src/web.rs
```diff
@@ -271,8 +271,8 @@ where
 }
 
 /// Wrapper for [`actix_web::web::resource`](https://docs.rs/actix-web/*/actix_web/web/fn.resource.html).
-pub fn resource(path: &str) -> Resource {
-    Resource::new(path)
+pub fn resource<S: AsRef<str>>(path: S) -> Resource {
```
wafflespeanut, line 274: We should stick to the signatures of `actix-web` functions rather than doing our own (for example, `web::resource` has a `T: IntoPattern` constraint). I dunno when we got diverged, but it's not good. Can you change this function to have that constraint? I'll check everything else and make the appropriate changes before release.

omid: That was my first try. But it's completely different!
The Resource is compeltely something else:

https://github.com/actix/actix-web/blob/master/src/resource.rs#L50

I can see similar code somewhere around actix-web 0.5! (didn't search a lot!)

omid:
The solution is to sync the existing actix implementation.

PR #227, opened 2020-09-25
do not include default features for actix-web

robjtede:
(kudos for using that version range 👀)


wafflespeanut (approved):

wafflespeanut:
Thanks!

strohel:
It seems that actix-web is referenced from another place that probably should receive similar treatment:
```
core/Cargo.toml-[dependencies]
core/Cargo.toml:actix-web = { version = ">=2.0,<4", optional = true }
```

It is also in `[dev-dependencies]` of top-level `Cargo.toml`, but that probably doesn't hurt as dev-dependencies are not propagated upstream.

wafflespeanut:
This is now fixed in master.

PR #228, opened 2020-09-26
Expose swagger ui

cfvescovo:
See https://github.com/wafflespeanut/paperclip/pull/174 for further information


plugins/actix-web/src/lib.rs
```diff
@@ -248,6 +251,24 @@ where
...
+    /// as a JSON with Swagger UI.
+    #[cfg(feature = "swagger-ui")]
+    pub fn with_swagger_ui_at(mut self, path: &str) -> Self {
+        self.inner = self.inner.take().map(|a| {
+            a.service(
+                actix_web::web::resource(path)
+                    .route(actix_web::web::get().to(SpecHandler(self.spec.clone()))),
+            )
+            .service(
+                actix_web::web::resource(path.to_owned() + "/swagger-ui.html")
```
wafflespeanut, line 265: This doesn't feel right. We need to think of a better way. I understand that the swagger UI needs access to the JSON spec, but we shouldn't mount the UI at some non-configurable path (what if users wanted this to be at `/api/swagger`?).

I propose to have a `path: Option<String>` field in the `App` wrapper type. We set this path to the JSON spec path (obtained when the user calls `with_json_spec_at` and we mount the UI handler only if this path is set, so if it's not set, we'll get a 404 from backend.

cfvescovo: Ok, I will add it

cfvescovo: I have made the changes you requested.

cfvescovo:
I have also updated the docs for our new function

cfvescovo:
@wafflespeanut Could you review the changes I have made?

wafflespeanut:
Hey, sorry I took long to take a look. As mentioned in the other PR, I don't like using `askama` for a simple path substitution. We already use `tinytemplate` for codegen (which has very few dependencies), and I'd prefer to use that instead. We can have a build script which loads the template at compile time and substitutes at runtime. If you do have time to work on it, then that'd be great. Otherwise, I'll take ownership of this PR :smile:

cfvescovo:
Ok, I will take a look at `tinytemplate`. If I manage to use it instead of `aksama, I will edit this PR accordingly.
Thank you for your feedback

cfvescovo:
I have just realised that I had forgot to edit this PR. As soon as I finish my implementation I will publish my changes.

cfvescovo:
@wafflespeanut I made the changes you requested.

cfvescovo:
Conflicts resolved!

Cargo.toml
```diff
@@ -64,7 +64,7 @@ actix2 = ["actix-base", "paperclip-actix/actix2"]
 actix3 = ["actix-base", "paperclip-actix/actix3"]
 actix-base = ["v2", "paperclip-macros/actix"]
 
-# OpenAPI support (v2 and codegen)
```
wafflespeanut, line 67: This line has been removed. You probably meant to move the feature on top of this line?

plugins/actix-web/Cargo.toml
```diff
@@ -19,10 +19,15 @@ paperclip-macros = { path = "../../macros", version = "0.4.0", features = ["acti
 parking_lot = ">=0.10,<0.12"
 serde_json = "1.0"
 once_cell = "1.4"
+tinytemplate = { version = "1.1.0", optional = true }
```
wafflespeanut, line 22: Let's relax the dependency to stick to `1.1`

plugins/actix-web/Cargo.toml
```diff
@@ -19,10 +19,15 @@ paperclip-macros = { path = "../../macros", version = "0.4.0", features = ["acti
...
+serde = { version = "1.0", optional = true }
+serde_qs = { version = "0.6", optional = true, features = ["actix"] }
```
wafflespeanut, line 24: This is probably a bad merge. We don't need `serde_qs` here as we already have it in `core`?

plugins/actix-web/Cargo.toml
```diff
@@ -19,10 +19,15 @@ paperclip-macros = { path = "../../macros", version = "0.4.0", features = ["acti
...
 
 [features]
 actix = ["actix3"]
 actix2 = ["actix-web2", "paperclip-core/actix2"]
 actix3 = ["actix-web3", "paperclip-core/actix3"]
 nightly = ["paperclip-core/nightly"]
 normalize = []
+swagger-ui = ["tinytemplate", "serde"]
+qs = ["serde_qs"]
```
wafflespeanut, line 33: Ditto on removing `qs` as we match `serde_qs` as a dependency feature.

plugins/actix-web/src/lib.rs
```diff
@@ -314,3 +351,40 @@ impl actix_web::dev::Factory<(), Ready<Result<HttpResponse, Error>>, Result<Http
...
+struct SwaggerUIHandler(String);
+
+#[cfg(feature = "swagger-ui")]
+#[derive(Serialize)]
+struct SwaggerUIContext {
+    api_spec_url: String,
+}
+
+#[cfg(feature = "swagger-ui")]
+static TEMPLATE: &str = include_str!("../templates/swagger-ui.html");
```
wafflespeanut, line 366: Let's move this from `plugins/actix-web/templates` to  `plugins/templates` so that it's applicable to other places as well.

plugins/actix-web/src/lib.rs
```diff
@@ -314,3 +351,40 @@ impl actix_web::dev::Factory<(), Ready<Result<HttpResponse, Error>>, Result<Http
...
+{
+    fn call(&self, _: ()) -> Ready<Result<HttpResponse, Error>> {
+        let mut tt = TinyTemplate::new();
+        let context = SwaggerUIContext {
+            api_spec_url: self.0.to_owned(),
+        };
+        // This is safe as we are using a given file as input so no errors can occur
+        let _ = tt.add_template("swagger-ui", TEMPLATE);
+
+        let rendered = tt.render("swagger-ui", &context);
```
wafflespeanut, line 380: Initializing this here will render it every time we get a request on that path. A more efficient way to do this would be to have a singleton or initialize it when we call `with_swagger_ui_at` and return `Result<Self, Error>` (which happens only once) and we store the rendered string in this struct.

cfvescovo:
@wafflespeanut I made the changes you requested.

plugins/actix-web/src/lib.rs
```diff
@@ -265,6 +280,53 @@ where
         self
     }
 
+    /// Exposes the previously built JSON specification with Swagger UI at the given path
+    ///
+    /// **NOTE:** you **MUST** call with_json_spec_at before calling this function
+    #[cfg(feature = "swagger-ui")]
+    pub fn with_swagger_ui_at(mut self, path: &str) -> Self {
```
wafflespeanut, line 287: We should be able to return `Result<Self, Error>` here, because missing spec and rendering error can be returned here as an error instead of serving them as a response.

cfvescovo: Ok

cfvescovo: Which error type should we return?

cfvescovo:
@wafflespeanut I am back!

cfvescovo:
I will have to resolve some conflicts though

plugins/actix-web/src/lib.rs
cfvescovo: @wafflespeanut let me know which Error type should be returned. Otherwise, I will return a generic Error value with a message describing what happened.

tiagolobocastro: We don't seem to have any library error for the moment. Use have `thiserror` so I would recommend adding a paperclip error type, and this will be the first error we have :)

patrickelectric:
@cfvescovo What is the path for the ui ? I did try `with_json_spec_at` path, and the same path with /ui in the end.

patrickelectric:
They ui appears to be broken
image

patrickelectric:
It appears to be missing:
swagger-ui.css.map

patrickelectric:
DevTools failed to load source map: Could not load content for http://0.0.0.0:8088/swagger-ui.css.map: HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE

cfvescovo:
It seems the template file is broken. You can find it at `plugins/template/swagger-ui.html`

patrickelectric:
Please check #331 as alternative

tiagolobocastro:
Closed in favour of #331

PR #232, opened 2020-09-30
Add actix-session support

### LINKED ISSUE #231: no function or associated item named `update_definitions` found for struct `actix_session::Session` in the current scope
Is there any way of fixing this error?

```
error[E0599]: no function or associated item named `update_definitions` found for struct `actix_session::Session` in the current scope
  --> src/main.rs:57:1
   |
57 | #[api_v2_operation]
   | ^^^^^^^^^^^^^^^^^^^ function or associated item not found in `actix_session::Session`
   |
   = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

omid:
Will fix #231

### DIFF
--- a/core/src/v2/actix.rs
+++ b/core/src/v2/actix.rs
@@ -277,6 +277,12 @@ impl OperationModifier for actix_multipart::Multipart {
     }
 }
 
+#[cfg(feature = "actix-session")]
+impl OperationModifier for actix_session::Session {
+    fn update_parameter(_op: &mut DefaultOperationRaw) {
+    }
+}
+
 macro_rules! impl_param_extractor ({ $ty:ty => $container:ident } => {
     #[cfg(feature = "nightly")]
     impl<T> Apiv2Schema for $ty {
--- a/core/src/v2/actix.rs
+++ b/core/src/v2/actix.rs
@@ -278,9 +278,7 @@ impl OperationModifier for actix_multipart::Multipart {
 }
 
 #[cfg(feature = "actix-session")]
-impl OperationModifier for actix_session::Session {
-    fn update_parameter(_op: &mut DefaultOperationRaw) {}
-}
+impl OperationModifier for actix_session::Session {}
 
 macro_rules! impl_param_extractor ({ $ty:ty => $container:ident } => {
     #[cfg(feature = "nightly")]
--- a/core/src/v2/actix.rs
+++ b/core/src/v2/actix.rs
@@ -38,7 +38,7 @@ pub trait OperationModifier: Apiv2Schema + Sized {
         update_security::<Self>(op);
     }
 
-    /// Update the security defition map (if needed).
+    /// Update the security definition map (if needed).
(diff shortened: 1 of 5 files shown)


core/src/v2/schema.rs
```diff
@@ -190,6 +190,12 @@ impl_type_simple!(
     DataType::File,
     DataTypeFormat::Binary
 );
+#[cfg(feature = "actix-session")]
+impl_type_simple!(
+    actix_session::Session,
+    DataType::File,
+    DataTypeFormat::Binary
+);
```
wafflespeanut, line 198: Let's do an empty impl instead? Because it doesn't feel right to have this schema for this type.

core/src/v2/actix.rs
```diff
@@ -277,6 +277,11 @@ impl OperationModifier for actix_multipart::Multipart {
     }
 }
 
+#[cfg(feature = "actix-session")]
+impl OperationModifier for actix_session::Session {
+    fn update_parameter(_op: &mut DefaultOperationRaw) {}
```
wafflespeanut, line 282: We can get rid of this function, because this is the default implementation. We should override `update_definitions` instead and have an empty function body.

wafflespeanut (changes requested):

core/src/v2/schema.rs
omid: What do you think to extend the macro to also support the following syntax?
I know it's doing nothing, but at least it's consistent!

```
impl_type_simple!(actix_session::Session);
```

core/src/v2/actix.rs
omid: Yep, sorry, missed that

core/src/v2/schema.rs
omid: I just did it... if you are unhappy with this, then I'll revert and do the impl separately.

core/src/v2/actix.rs
```diff
@@ -277,6 +277,9 @@ impl OperationModifier for actix_multipart::Multipart {
     }
 }
 
+#[cfg(feature = "actix-session")]
+impl OperationModifier for actix_session::Session {}
```
wafflespeanut, line 281: We have to override `update_definitions` to an empty function

core/src/v2/schema.rs
wafflespeanut: No, that's okay. It's good to have it. :smile:

core/src/v2/actix.rs
omid: So both `update_security_definitions` and `update_definitions` I think.

wafflespeanut: No, just `update_definitions` would be good, because we haven't implemented any security stuff for that type.

omid: done

wafflespeanut (approved):
Thanks!

PR #237, opened 2020-10-01
Remove object type from json and yaml Values

### LINKED ISSUE #235: `serde_json::Value` should show as unknown type
`serde_json::Value` currently specs as an object, although it can be any kind of json type.

Would this be more correct to have no type?
```
-impl_type_simple!(serde_json::Value, DataType::Object);
-impl_type_simple!(serde_yaml::Value, DataType::Object);
+
+impl Apiv2Schema for serde_json::Value {}
+impl Apiv2Schema for serde_yaml::Value {}
+
```

platy:
I'm not sure if this is more correct, but it feels more correct to me. Out of all of the variants of `serde_json::Value` only one of them is an `object`.

Closes #235


wafflespeanut:
You're right. I dunno how it got there. `Value` types are actually "any" types and in OpenAPI, any type has empty schema.

wafflespeanut (approved):
Thanks!

Issue #241, opened 2020-10-06
How can I return 204 No Content HTTP status code?


omid:
What should I use for a code like this?

```
#[api_v2_operation]
pub async fn delete(id: Path<Uuid>) -> MyResult<()> {
    // delete the entity

    Ok(())
}
```

If I do like the following code, the reponse section of the swagger is empty:
```
#[api_v2_operation]
pub async fn delete(id: Path<Uuid>) -> MyResult<HttpResponse> {
    // delete the entity

    Ok(HttpResponse::NoContent().finish())
}
```

platy:
This is related to #62, I also needed to do status codes other than 200 recently, including no content. Unfortunately you can't use HttpResponse as it doesn't give any type information for paperclip to use. I provided the solution I used in a gist (see below). I'm not sure whether to make a PR to add this as I'm not sure if it covers use cases generally enough.

https://gist.github.com/platy/e0bc44e0ce91c8268dcf0d9689e1f974#file-status_codes-rs-L91-L125

If you include the `NoContent` struct from there you can return `Ok(NoContent)` from your `fn() -> MyResult<NoContent>` and both actix-web and paperclip should behave as expected.

omid:
Thanks @platy, your gist code is what I need.

I would say you can make a pull request, then if anybody else wants to add a new status code, or there is something missing, we can improve it. It shouldn't be perfect from the first place.

platy:
Will do, thanks

platy:
@omid I raised a PR #243, also I just noticed that not only are we coming across the same issues at the same time while integrating paperclip, we're also based in the same city 😆

omid:
@platy lol, I can see Berlin is changing :laughing:

omid:
Kudos to @platy, it's there in the master branch already!

PR #243, opened 2020-10-06
Add types for encoding successful response codes

platy:
Add types to encode successful response codes in actix-web:
* `NoContent`
* `AcceptedJson`
* `CreatedJson`

Allows status codes to be used both by paperclip and actix-web, eg.
```rust
use paperclip::actix::web::Json;
use paperclip::actix::{CreatedJson, AcceptedJson, NoContent};

// 201 Created 
#[api_v2_operation]
async fn adopt_pet(body: Json<Pet>) -> Result<CreatedJson<Pet>, ()> {
    let pet: Pet = body.into_inner();
    // bring the pet home
    Ok(CreatedJson(pet))
}
// 204 No Content
#[api_v2_operation]
async fn acknowledge_pet(body: Json<Pet>) -> NoContent {
    NoContent
}
```

I was not really sure where to put this, I put it in `paperclip_core::v2::actix` because it fit there the easiest.

#62 #241

### DIFF
--- a/core/src/v2/actix.rs
+++ b/core/src/v2/actix.rs
@@ -5,16 +5,20 @@ use super::models::{
 #[cfg(feature = "actix-multipart")]
 use super::schema::TypedData;
 use super::schema::{Apiv2Errors, Apiv2Operation, Apiv2Schema};
+use crate::util::{ready, Ready};
 use actix_web::{
+    http::StatusCode,
     web::{Bytes, Data, Form, Json, Path, Payload, Query},
-    HttpRequest, HttpResponse, Responder,
+    Error, HttpRequest, HttpResponse, Responder,
 };
 use pin_project::pin_project;
 
+use serde::Serialize;
 #[cfg(feature = "serde_qs")]
 use serde_qs::actix::QsQuery;
 
 use std::collections::BTreeMap;
+use std::fmt;
 use std::future::Future;
 use std::pin::Pin;
 use std::task::{Context, Poll};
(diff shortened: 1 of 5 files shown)


platy:
I decided to combine the status code with the Json wrapper because it makes the usage shorter, but maybe it makes sense to separate them and leave it up to the user to define types and functions which make the usage better.

wafflespeanut:
This would certainly be the way in the previous version of papercliop, but now we're generating the code for each handler, so we can solve this in an easier way through the macro attribute :smile: 
```
#[api_v2_operation(status(code = 202, description = "I accept"))]  // we can support multiple statuses too
fn my_handler () -> Foo { /* */ }
```
What do you think?

platy:
I really appreciate the macro attribute for being able to describe what the status code means for that particular handler, but I appreciate the return type for joining the behaviour with the spec and meaning that the spec can't be wrong.

core/src/v2/actix.rs
```diff
@@ -521,3 +525,120 @@ where
         new.update_definitions(name, map);
     }
 }
+
+macro_rules! json_with_status {
+    ($name:ident => $status:expr) => {
+        pub struct $name<T: Serialize + Apiv2Schema>(pub T);
```
wafflespeanut, line 531: Let's get rid of the constraints in the type. I prefer having it in the impl instead. Also, let's have a default type.
```rust
pub struct $name<T = ()>(pub T);
```

core/src/v2/actix.rs
```diff
@@ -521,3 +525,120 @@ where
...
+            }
+        }
+    };
+}
+
+json_with_status!(CreatedJson => StatusCode::CREATED);
+json_with_status!(AcceptedJson => StatusCode::ACCEPTED);
+
+#[derive(Debug)]
+pub struct NoContent;
```
wafflespeanut, line 611: Do we have to implement this manually? I understand it's because `NoContent` is not supposed to return anything, but still, we can use `json_with_status!` for `NoContent` too.

platy: I'm for making this DRY, but I think if the behaviour of `NoContent` was serialising `()` as json it would respond with `null`. So i'd need to make the approach more generic and make it apply to any responder rather than reimplementing Json. I like it but I'm not sure how to do it yet

wafflespeanut (approved):
Alright, let's keep this as is for now.

PR #268, opened 2020-11-26
Add Json and Yaml types

omid:
The current generated code for Jsons and Yamls are wrong. Since there is no specifc type for these formats in OpenApi, they should be "string". (Am I right?)

So for the following struct:

```
struct Pet {
    name: String,
    detail: Option<Jsonb>,
    id: Option<i64>,
}
```

Instead of this (notice the _detail_ field):

```
{
	"definitions": {
		"Pet": {
			"type": "object",
			"properties": {
				"detail": {},
				"id": {
					"type": "integer",
					"format": "int64"
				},
				"name": {
					"type": "string"
				}
			},
			"required": ["name"]
		}
	}
}
```

It will generate this:
```
{
	"definitions": {
		"Pet": {
			"type": "object",
			"properties": {
				"detail": {
					"type": "string"
				},
				"id": {
					"type": "integer",
					"format": "int64"
				},
				"name": {
					"type": "string"
				}
			},
			"required": ["name"]
		}
	}
}
```

### DIFF
--- a/core/src/v2/schema.rs
+++ b/core/src/v2/schema.rs
@@ -185,6 +185,8 @@ impl_type_simple!(isize, DataType::Integer, DataTypeFormat::Int64);
 impl_type_simple!(u64, DataType::Integer, DataTypeFormat::Int64);
 impl_type_simple!(u128, DataType::Integer, DataTypeFormat::Int64);
 impl_type_simple!(usize, DataType::Integer, DataTypeFormat::Int64);
+impl_type_simple!(serde_json::Value, DataType::String);
+impl_type_simple!(serde_yaml::Value, DataType::String);
 
 #[cfg(feature = "actix-multipart")]
 impl_type_simple!(
@@ -287,8 +289,6 @@ pub trait Apiv2Schema {
 }
 
 impl Apiv2Schema for () {}
-impl Apiv2Schema for serde_json::Value {}
-impl Apiv2Schema for serde_yaml::Value {}
 
 impl<T: TypedData> Apiv2Schema for T {
     fn raw_schema() -> DefaultSchemaRaw {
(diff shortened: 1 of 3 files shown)


core/src/v2/schema.rs
```diff
@@ -287,8 +289,6 @@ pub trait Apiv2Schema {
 }
 
 impl Apiv2Schema for () {}
```
omid, line 291: I'm not sure about this one also.
Don't see any use case and also in the other hand, the generated OpenApi json may be wrong!

There is also no test for this line.

dunnock:
Actually when you output Value it usually means it is embedded object. You rarely serialize json into string in the output in response type

omid:
@dunnock, I think then in that case, it will refer to a new object (struct). Somehting like:

```
struct Pet {
    name: String,
    detail: Option<Detail>,
    id: Option<i64>,
}

struct Detail {
    something: String,
}
```

core/src/v2/schema.rs
```diff
@@ -185,6 +185,8 @@ impl_type_simple!(isize, DataType::Integer, DataTypeFormat::Int64);
 impl_type_simple!(u64, DataType::Integer, DataTypeFormat::Int64);
 impl_type_simple!(u128, DataType::Integer, DataTypeFormat::Int64);
 impl_type_simple!(usize, DataType::Integer, DataTypeFormat::Int64);
+impl_type_simple!(serde_json::Value, DataType::String);
+impl_type_simple!(serde_yaml::Value, DataType::String);
```
dunnock, line 189: This does not conform to actix_web Responder implementation as Value does not implement String Responder. 

You would usually wrap Value as `web::Json<Value>` and it will output object. Same when Value is used as part of a Serializable struct - it will use wrapper's implementation.

If you need to output Value or your strust as a String you would usually need to implement your JsonAsString wrapper, where you can define this mapping and implement custom Apiv2Schema.

omid: Thanks. Got it.

PR #269, opened 2020-11-27
Transform api operations based upon middleware wrappers

platy:
I came across the need to modify the api spec based on properties of middleware transformers. In our case it was RBAC middleware which specified different roles for different endpoints.

This change adds another trait to the parameter on the `Resource` wrapper, `TransformApi`. In this use case it allowed the roles required on the endpoints to be automatically added to `operation.tags` and into `operation.description`, which seems like it will be very useful.

I have a couple of concerns about it though:
* this changes the signature of `wrap` so it is a breaking change for users of that function - it could be given a new name
* I've made `RouteWrapper` public - maybe it should use a different type?
* the transformers need to be cloned to be passed into the inner and kept in the wrapper - but in my case they are very small
* the transformers are being kept in the wrapper in a type-system linked list, this avoided allocations and is fine in my case but maybe others would have a problem with it

### DIFF
--- a/plugins/actix-web/src/web.rs
+++ b/plugins/actix-web/src/web.rs
@@ -35,12 +35,13 @@ const METHODS: &[Method] = &[
 /* Resource */
 
 /// Wrapper for [`actix_web::Resource`](https://docs.rs/actix-web/*/actix_web/struct.Resource.html)
-pub struct Resource<R = actix_web::Resource> {
+pub struct Resource<R = actix_web::Resource, N = ()> {
     path: String,
     operations: BTreeMap<HttpMethod, DefaultOperationRaw>,
     definitions: BTreeMap<String, DefaultSchemaRaw>,
     security: BTreeMap<String, SecurityScheme>,
     inner: R,
+    api_transformations: N,
 }
 
 impl Resource {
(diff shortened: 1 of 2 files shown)


dunnock:
BTW, would not the roles required on endpoint better managed with Apiv2Security derive? 
We are managing this internally via FromRequest handlers on a User and AccessToken, that seems like better fit in actix world. Not sure about your use case details though.

In regards to back compatibility - we are doing 2 phase initialization of actix-web, first is for all services which we exclude from schema, then doing wrap_api and all schema based initializations. I guess this can be adviced as a migration approach if your PR would be accepted.

platy:
Yes I think it might be a nice solution to have a security parameter which contained type information about the roles that are allowed access and to implement the schema documentation there. But for now that seems to big of a change for this project. The role information is already there in the middleware and available at runtime, it seems reasonable that middleware could alter the spec as it could change anything about the handling of resources.

I have only worked on this one actix web project so I don't know anything about how they generally look, but I feel that more extensibility would be helpful to paperclip.

dunnock:
There is a different way to pass data from middleware now `ReqData` so that might be a better way to modify the schema via wrapped types? https://github.com/wafflespeanut/paperclip/issues/300

platy:
@dunnock Thank's for the info - I'll keep that in mind if we update to actix-web 3 at some point

PR #270, opened 2020-11-27
Add NaiveTime type for actix plugin

platy:
Add the last of `chrono`s dat and time types.
The spec doesn't have a format for time so this is just an unformated string type.


wafflespeanut (approved):
Thanks!

PR #271, opened 2020-11-27
Preserve newlines in docstring description

sfisol:
Hello!

This small change adds "\n" at the end of every line in description gathered from docstring. This allows to display markdown when using swagger.

Whereas this works perfectly for me, I'm not sure about other use-cases. For example if in future paperclip will support YAML format then this could potentially break something.

If you have any advice let me know and I'll try to amend this.

### DIFF
--- a/macros/src/actix.rs
+++ b/macros/src/actix.rs
@@ -343,7 +343,7 @@ fn extract_fn_documentation(
         }
     };
     let summary = none_if_empty(summary.into_iter().collect::<String>().trim());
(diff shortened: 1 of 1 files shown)


amarjanica:
Hi, I have the same issue. @sfisol I'm curious, why not just write `description.join("\n").trim()`?

sfisol:
You're right @amarjanica, I blindly followed the original pattern :-) Amended. Thanks!

wafflespeanut (approved):

PR #272, opened 2020-12-05
Fix optional type aliases

### LINKED ISSUE #257: Actix: type alias makes Option<T> required
Having a schema like this:
```rust
type Email = Option<String>;

#[derive(Apiv2Schema, Serialize)]
struct User {
    id: i32,
    username: String,
    email: Email,
}
```

The `email` field is optional, but in the generated spec it is shown as required (`"required":["email","id","username"]`)

Full code:
```rust
use actix_web::{App, HttpServer};
use paperclip::actix::web::{self, Json};
use paperclip::actix::{api_v2_operation, Apiv2Schema, OpenApiExt};
use serde::Serialize;

type Email = Option<String>;

#[derive(Apiv2Schema, Serialize)]
struct User {
    id: i32,
    username: String,
    email: Email,
}

#[api_v2_operation]
async fn example() -> Json<User> {
    Json(User {
        id: 1,
        username: "john".into(),
        email: None,
    })
}

#[actix_web::main]
async fn main() -> std::io::Result<()> {
    HttpServer::new(|| {
        App::new()
            .wrap_api()
            .route("/", web::get().to(example))
            .with_json_spec_at("/api/spec")
            .build()
    })
    .bind("127.0.0.1:8080")?
    .run()
    .await
}
```

Cargo.toml:
```toml
[dependencies]
actix-web = "3"
paperclip = { git = "https://github.com/wafflespeanut/paperclip", features = ["actix"] }
serde = { version = "1.0", features=["derive"] }
```

omid:
Fix #257

### DIFF
--- a/core/src/v2/schema.rs
+++ b/core/src/v2/schema.rs
@@ -286,6 +286,10 @@ pub trait Apiv2Schema {
     fn security_scheme() -> Option<SecurityScheme> {
         None
     }
+
+    fn is_required() -> bool {
+        true
+    }
 }
 
 impl Apiv2Schema for () {}
@@ -312,6 +316,10 @@ impl<T> Apiv2Schema for Option<T> {
     default fn security_scheme() -> Option<SecurityScheme> {
         None
     }
+
+    fn is_required() -> bool {
+        false
+    }
 }
 
 impl<T: Apiv2Schema> Apiv2Schema for Option<T> {
@@ -323,6 +331,10 @@ impl<T: Apiv2Schema> Apiv2Schema for Option<T> {
 
     fn security_scheme() -> Option<SecurityScheme> {
         T::security_scheme()
+    }
+
+    fn is_required() -> bool {
+        false
     }
 }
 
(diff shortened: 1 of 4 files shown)


macros/src/actix.rs
```diff
@@ -769,37 +769,23 @@ fn add_optional_impl(name: &Ident, generics: &Generics) -> proc_macro2::TokenStr
...
-        Type::Path(ref p) => {
-            let ty = p
-                .path
-                .segments
-                .last()
-                .expect("expected type for struct field");
-
-            if p.path.segments.len() == 1 && &ty.ident == "Option" {
-                is_required = false;
-            }
```
wafflespeanut, line 784: This is certainly not what I expected in code. I guess you can't always remember how the code evolves.

core/src/v2/schema.rs
```diff
@@ -312,6 +316,10 @@ impl<T> Apiv2Schema for Option<T> {
     default fn security_scheme() -> Option<SecurityScheme> {
         None
     }
+
+    fn is_required() -> bool {
```
wafflespeanut, line 320: Can we make this a constant? It's a primitive value after all.

omid: Done :)

wafflespeanut:
One last thing. Can you add/modify a test for it?

wafflespeanut (approved):
Thanks!

PR #278, opened 2020-12-17
Define list items on query parameters

### LINKED ISSUE #239: array types in `Query` don't emit items

platy:
Fixes #239.

For paramters with sequences this change also adds the .items parameter of the ParameterObject and is generated recursively.

In #242 I also tried to correctly document `serde_qs::QsQuery`, but I now don't think it is possible to correctly document `serde_qs` sequences in openapi v2 and am closing that PR.

### DIFF
--- a/core/src/v2/actix.rs
+++ b/core/src/v2/actix.rs
@@ -2,8 +2,8 @@
 use super::schema::TypedData;
 use super::{
     models::{
-        DefaultOperationRaw, DefaultResponseRaw, DefaultSchemaRaw, Either, Parameter, ParameterIn,
-        Response, SecurityScheme,
+        DefaultOperationRaw, DefaultResponseRaw, DefaultSchemaRaw, Either, Items, Parameter,
+        ParameterIn, Response, SecurityScheme,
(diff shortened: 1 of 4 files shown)


wafflespeanut (approved):
Thank you for this!

PR #280, opened 2020-12-22
Dump spec from actix configuration with spec() method on App builder

### LINKED ISSUE #279: How to generate offline schema from actix definitions?
Hi, I'm a new user to actix and paperclip (and rust to a degree). I've written a small service that carries out and dumps a swagger.json on HTTP request.

While attempting to generate the client for it, I came across a conundrum. How do I generate it without manually starting the service and requesting it? I'd like to be able to dump it to stdout, basically.

Is this possible? If not, is it Ok to patch in?

erikh:
closes #279 

This allows you to return a serde_json::Value from the app builder, allowing you to dump a definition at any time.

Signed-off-by: Erik Hollensbe <github@hollensbe.org>

### DIFF
--- a/plugins/actix-web/src/lib.rs
+++ b/plugins/actix-web/src/lib.rs
@@ -279,6 +279,11 @@ where
         call(self, spec)
     }
 
+    // Returns the spec as a serde value. Does not continue to build the application.
+    pub fn spec(self) -> serde_json::Value {
+        serde_json::to_value(&*self.spec.read()).expect("generating json spec")
+    }
+
     /// Builds and returns the `actix_web::App`.
     pub fn build(self) -> actix_web::App<T, B> {
         self.inner.expect("missing app?")


plugins/actix-web/src/lib.rs
```diff
@@ -279,6 +279,11 @@ where
         call(self, spec)
     }
 
+    // Returns the spec as a serde value. Does not continue to build the application.
+    pub fn spec(self) -> serde_json::Value {
+        serde_json::to_value(&*self.spec.read()).expect("generating json spec")
+    }
```
wafflespeanut, line 285: Why can't we make use of `with_raw_json_spec` method? It passes the built JSON spec as `serde_json::Value`, which can be used for custom mounting and inspection, no?

erikh: Yes, but unfortunately for situations like printing to stdout I still need to return something in that closure as I understand it: https://paperclip.waffles.space/paperclip/actix/struct.App.html#method.with_raw_json_spec -- this was the barrier I was running into. The result with this patch is now that my code looks something like this for my `dump-schema` subcommand:

```rust
pub fn dump_schema() -> std::io::Result<()> {
    std::io::stdout().lock().write_all(
        App::new()
            .wrap_api()
            .configure(configure)
            .spec()
            .to_string()
            .as_bytes(),
    )
}
```

Sorry if there's a better way of doing this and I'm just not aware of it.

dunnock: Maybe not looking that good but farily good for a CLI too, you could do it now like
```
use actix_web::test;
let mut app = test::init_service( || 
      App::new()
            .wrap_api()
            .configure(configure)
            .with_json_spec_at("/swagger")
            .build()
).await;
let schema: Value = test::read_response_json(&mut app, test::TestRequest::with_uri("/swagger"));
```

dgr113: Why can't I just open access to `spec` field of App object (with using getter) ?
In my opinion the only problem is that this is a private field.
I solved problem with a test query, but it's ... ugly!

dunnock: This is not a common requirement and has at least 2 ways to implement on a client side - via `with_raw_json_spec` and `test::init_service`.

erikh: hey sorry for the incredibly late response. I've mostly moved past this problem in my project, not really willing to carry a torch for this patch.

If ya'll have a use for it, great, but if not, you won't upset me by closing it. Hope that's reasonable enough and nobody is upset.

PR #281, opened 2020-12-22
Provide the --no-root flag to the CLI which allows the crate to be embedded

erikh:
This omits the workspace definition, allowing it to be embedded in other
repositories.

hope this patch is welcome. seems to be doing what i need locally, but I may have missed something.

Signed-off-by: Erik Hollensbe <github@hollensbe.org>

### DIFF
--- a/src/v2/codegen/mod.rs
+++ b/src/v2/codegen/mod.rs
@@ -65,7 +65,7 @@ pub struct CrateMeta {
     /// Whether we're planning to emit a lib, app or module.
     pub mode: EmitMode,
     /// Whether or not to make this a root crate.
-    pub root: bool,
+    pub no_root: bool,
     // Marker to avoid potential breakage when more public fields come in.
     _marker: (),
 }
--- a/src/v2/codegen/mod.rs
+++ b/src/v2/codegen/mod.rs
@@ -64,6 +64,8 @@ pub struct CrateMeta {
     pub authors: Option<Vec<String>>,
     /// Whether we're planning to emit a lib, app or module.
     pub mode: EmitMode,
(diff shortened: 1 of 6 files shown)


wafflespeanut (approved):
Good idea! Thanks!

wafflespeanut:
I see that the tests have failed. Feel free to add/modify them accordingly.

erikh:
Sorry about that. I've provided a version now that should pass tests, I'll keep an eye on it.

I'll also add a test that omits the workspace, assuming I can figure this out -- still new to rust.

erikh:
Here's some tests. Thanks for letting me poke in your pie!

src/v2/codegen/mod.rs
```diff
@@ -64,6 +64,8 @@ pub struct CrateMeta {
     pub authors: Option<Vec<String>>,
     /// Whether we're planning to emit a lib, app or module.
     pub mode: EmitMode,
+    /// Whether or not to make this a root crate.
+    pub root: bool,
```
wafflespeanut, line 68: Can we just make this `no_root`? Because, this would otherwise be a breaking change for `EmitterState::default()`, as the default for `root` would be `false`.

erikh: latest version has this change. Let me know if you'd like the commits squashed.

wafflespeanut (approved):
Good work! Thanks!

src/v2/codegen/mod.rs
wafflespeanut: Nah, it's alright, I'll squash it in merge :smile:

PR #287, opened 2021-01-19
fix(tuples): make doc work in tuple struct path

tiagolobocastro:
Add missing schema to unnamed structs with >1 element.
Fix issue with unnamed structs of 1 element caused by:
6429ef4
Fix usage of structs with Path with tuples: Path<(A,B,C)>

Unnamed struct element now get named starting from 0,1..... so that we
can keep the list of properties, though they should not really be used
anywhere else than a Path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Optional type aliases like `type Email = Option<String>` will not be added to the `required` fields.
+- Actix plugin: Path tuples now inherit field names and descriptions from doc comments
 
 ## [0.5.0] - 2020-11-28
 ### Added
--- a/core/src/v2/actix.rs
+++ b/core/src/v2/actix.rs
@@ -321,10 +321,10 @@ macro_rules! impl_param_extractor ({ $ty:ty => $container:ident } => {
                     data_type: def.data_type,
                     format: def.format,
                     enum_: def.enum_,
+                    description: def.description,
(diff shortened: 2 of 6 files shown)


dunnock (commented):
That looks good, can you please extend test to see if it's handling struct with 2 unnamed arguments? It's handled a bit differently and produces `object` now, would be great if that inherits description too

tiagolobocastro:
> That looks good, can you please extend test to see if it's handling struct with 2 unnamed arguments? It's handled a bit differently and produces `object` now, would be great if that inherits description too

Yeah it produces object probably because unnamed structs with >1 elem are left with an empty schema.
I don't know if it's possible to improve things but I'll give it a go.

tiagolobocastro:
Ok so we can get unnamed structs working by naming them such as they sort of are in rust, with names starting with 0, much like you'd do self.0?
That works pretty well for a Uri Path!
```
struct Test(String, String);
fn test(web::Path(test): web::Path<Test>) ...
```
Less so for a Body because you'd see the openapi scheme as:
```
{
  "0": "string",
  "1": "string"
}
```
But, that's a user choice I suppose... 
I'll push this up once I have time to tidy it up and write up the tests for this.

tiagolobocastro:
@dunnock I've pushed, please let me know what you think.

dunnock (approved):
That's awesome! Sorry it took a while to dig into that

tiagolobocastro:
> That's awesome! Sorry it took a while to dig into that

no worries, thank you!

wafflespeanut:
Thank you for working on this!

PR #289, opened 2021-01-24
feat(macro): add macro helpers for method routing

tiagolobocastro:
Add macro helpers for routing methods get, put, post, delete:
these are based on the actix_web macros and intended to be facilitate
use paperclip on actix_web code which already makes use of the original
macros from actix_web:
https://docs.rs/actix-web-macros/0.1.0/actix_web_macros/

### DIFF
--- a/tests/test_app.rs
+++ b/tests/test_app.rs
@@ -13,8 +13,8 @@ use futures::future::{ok as fut_ok, ready, Future, Ready};
 use once_cell::sync::Lazy;
 use paperclip::{
     actix::{
-        api_v2_errors, api_v2_operation, web, Apiv2Schema, Apiv2Security, CreatedJson, NoContent,
-        OpenApiExt,
+        api_v2_errors, api_v2_operation, delete, get, post, put, web, Apiv2Schema, Apiv2Security,
+        CreatedJson, NoContent, OpenApiExt,
     },
     v2::models::{DefaultApiRaw, Info, Tag},
 };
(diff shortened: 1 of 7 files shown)


tiagolobocastro:
Please let me know what you think of this!
Also, could I please get some help getting the doc_test example to compile? Had to comment it out because it failed to compile, I can't get to find the actix_web or paperclip crate :(

tests/test_app.rs
```diff
@@ -2571,17 +2571,192 @@ fn test_security_app() {
     );
 }
 
+#[test]
+fn test_method_macro() {
+    #[get("/v0/pets")]
+    #[api_v2_operation]
```
dunnock, line 2577: That's neat! The only thing I'm wondering is if `api_v2_operation` macro is not defined here I guess we should pretend it is...

tiagolobocastro: Yeah the `route` (get/put...) macros could automatically add the `#[api_v2_operation]`, not exactly sure how to do that yet but it's probably possible!
Oh and what about adding the `route` to the api macro, as such:
`#[api_v2_operation(get("/v0/pets"))]` ?
Though the downside is more "labour" converting things. Any thoughts?

dunnock: Better to keep compatibility with established `actix_web_macros` definitions, main idea is to make a transition to use paperclip as easy as possible.

tiagolobocastro: Yeah agreed - I'll see if I can detect the existence of the `api_v2_operation` :+1:

tiagolobocastro: Sadly I do not know how to do this :( 
I tried using syn::parse_outer which parses a macro just fine, but because the item also has the function it fails...

dunnock: I think that's fine , by itself this change is valuable as enables paperclip for usage with actix-web-macros

dunnock (approved):
I think this change is valuable, enabling usage of actix-web-macros, thanks @tiagolobocastro . LGTM!

PR #297, opened 2021-02-11
fix(doc): missing documentation on types

### LINKED ISSUE #290: Missing Documentation for Type
I noticed on some of the tests, the Pet description was missing! Eg:
> "Pet": {
>           **"description": "Pets are awesome!",**
>           "properties": { ...
> }

I've managed to reproduce it in this simple code:
https://github.com/tiagolobocastro/paperclip/commit/5b18ddf3dbaff1c39058e7e617afdd0421bb466b
Now, if you change the return of _echo_pet_ from Pet into **`Vec<Pet>`** then the description appears!

tiagolobocastro:
Add missing documentation to struct definitions.
Updated tests.

Resolves  #290

### DIFF
--- a/macros/src/actix.rs
+++ b/macros/src/actix.rs
@@ -491,7 +491,9 @@ pub fn emit_v2_definition(input: TokenStream) -> TokenStream {
                 schema.data_type = Some(DataType::Object);
             ));
             match &s.fields {
-                Fields::Named(ref f) => handle_field_struct(f, &props, &mut props_gen),
+                Fields::Named(ref f) => {
+                    handle_field_struct(f, &item_ast.attrs, &props, &mut props_gen)
+                }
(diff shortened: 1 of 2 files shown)


tiagolobocastro:
Mind having a look at this @dunnock? Thanks

Btw also noticed that Body type parameters don't have a description (it's set to None). Should they?

dunnock (approved):
Really good spot, LGTM! 

Body might have description according to the spec, but I don't think we have any place where we can set it.. I did not see a practice of doc comments for function parameters. On a bright side - most tools will display description of referenced parameter. Though @wafflespeanut afaik there was intent to allow such extensions as api_v2_operation macros attributes?

tiagolobocastro:
If this one and 289 go in, do you think we can have a new release? Would be nice to pick these up without patching the toml to point to github.

PR #301, opened 2021-02-15
support actix-files static file serving

Tarnadas:
This PR adds support for actix_files::Files.

It will just make it compile and won't add any Swagger definitions, since static files that are served should not have any impact on the REST API.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,7 +72,7 @@ v2 = ["paperclip-macros/v2", "paperclip-core/v2"]
 # Features for implementing traits for dependencies.
 actix-multipart = ["paperclip-core/actix-multipart"]
 actix-session = ["paperclip-core/actix-session"]
-actix-files = ["paperclip-core/actix-files"]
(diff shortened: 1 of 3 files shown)


dunnock:
This one is not necessary, as paperclip does not demand all services to be paperclip compatible. To make paperclip working it provides trait extensions marking start of the paperclip compatible services via `App::wrap_api()` and end of that section via `App::build_with_json_spec_at()`.  For example this is taken from real production setup:
```
          app.configure(routing::routes::<B>)
              .wrap_api_with_spec(crate::swagger::default_spec())
              .configure(routing::swagger_routes::<B>)
              .with_json_spec_at("/swagger")
              .build()
              .default_service(web::get().to(|| HttpResponse::NotFound().json(json!({"error": "Not found"}))))
```

The problem here is that if we take approach of extending paperclip, there are too many crates out there which need to be covered, so would rather support only those which are required indeed.

Tarnadas:
thanks for clarifying

PR #307, opened 2021-02-21
feat(error responses): error responses definitions

### LINKED ISSUE #306: Use different response types
So, looks like openapi supports using a schema for error types, including different schemas for each error code.

See: https://swagger.io/docs/specification/2-0/describing-responses/

I can see we can add error descriptions as per: https://paperclip.waffles.space/actix-status-codes.html

Looking at the code doesn't look like we can do it but I'll ask anyway in case I missed it:
Is it possible to add a type to the error?
Can we specify a type for each status code?

tiagolobocastro:
Ability to specify an error type schema to an error response, either to
 each specific error code or a common for all.
Add error overlay over an existing Apiv2Error which allows us to filter
 out errors from the existing type which may not suit a particular
 URI handler.
Closes #306

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for the `#[serde(skip)]` attribute in structs and enums.
 
 ### Changed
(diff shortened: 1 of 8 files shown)


tiagolobocastro:
any thoughts @dunnock? Sorry I keep leaning on you, not sure if there's anyone else active here?

PR #309, opened 2021-03-16
feat(skip): add support for serde skip

### LINKED ISSUE #305: Skip struct fields
Hi, I have a struct with my own serializer/deserializer and so I have had to resort to implement the ApiV2Scheme by hand to remove the inner types of my struct, eg:

```rust
/// Generic JSON value eg: { "size": 1024 }
#[derive(Debug, Clone)]
pub struct JsonGeneric {
    inner: serde_json::Value,
}
impl paperclip::actix::OperationModifier for JsonGeneric {}
impl paperclip::v2::schema::Apiv2Schema for JsonGeneric {
    const DESCRIPTION: &'static str =
        "Generic JSON value eg: { \"size\": 1024 }";
    fn raw_schema() -> paperclip::v2::models::DefaultSchemaRaw {
        use paperclip::v2::models::{DataType, DefaultSchemaRaw};
        let mut schema = DefaultSchemaRaw::default();
        schema.data_type = Some(DataType::Object);
        schema.name = Some("JsonGeneric".into());
        schema.description = Some(Self::DESCRIPTION.to_string());
        schema
    }
}

impl Serialize for JsonGeneric {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: serde::Serializer,
    {
        self.inner.serialize(serializer)
    }
}
impl<'de> Deserialize<'de> for JsonGeneric {
    fn deserialize<D>(deserializer: D) -> Result<JsonGeneric, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        let value = serde_json::Value::deserialize(deserializer)?;
        Ok(JsonGeneric::from(value))
    }
}
```

Is there anyway to "skip" fields, much like the serde skip attribute?

tiagolobocastro:
Skips fields if the serde skip attribute is found.

Closes #305

### DIFF
--- a/macros/src/actix.rs
+++ b/macros/src/actix.rs
@@ -801,16 +801,30 @@ fn handle_unnamed_field_struct(
             let docs = extract_documentation(struct_attr);
             let docs = docs.trim();
 
-            props_gen.extend(quote!({
-                let mut s = #ty_ref::raw_schema();
-                if !#docs.is_empty() {
-                    s.description = Some(#docs.to_string());
-                }
-                schema = s;
-            }));
+            if SerdeSkip::exists(&field.attrs) {
+                props_gen.extend(quote!({
+                    let mut s: DefaultSchemaRaw = Default::default();
+                    if !#docs.is_empty() {
+                        s.description = Some(#docs.to_string());
+                    }
+                    schema = s;
+                }));
+            } else {
+                props_gen.extend(quote!({
+                    let mut s = #ty_ref::raw_schema();
+                    if !#docs.is_empty() {
+                        s.description = Some(#docs.to_string());
+                    }
+                    schema = s;
+                }));
+            }
         }
     } else {
         for (inner_field_id, field) in (&fields.unnamed).into_iter().enumerate() {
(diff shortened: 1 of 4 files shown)


macros/src/actix.rs
```diff
@@ -1093,6 +1115,42 @@ impl SerdeRename {
...
+/// Serde skip (https://serde.rs/variant-attrs.html)
+/// Never serialize or deserialize this variant.
+/// There are other variants available (skip_serializing,skip_deserializing) though it's not clear
+/// how this should be handled since we use the same Schema for Ser/DeSer
+struct SerdeSkip;
+
+impl SerdeSkip {
+    /// Traverses the field attributes and returns whether the field should be skipped or not
+    /// dependent on finding the `#[serde(skip]` attribute.
+    fn from_field_attrs(field_attrs: &[Attribute]) -> bool {
```
omid, line 1127: `exists` is a better name, like what we have in the `SerdeFlatten` struct.

tiagolobocastro: it is indeed, I've fixed it

PR #312, opened 2021-03-23
fix: add missing Configure app_data and trim paths

tiagolobocastro:
Trimming the base path is useful when using scopes otherwise you could
 get the base uri repeated, eg: "/v0/v0". Obviously this depends on the
 use case so call only if you want this.


tomsik68:
The `app_data` change looks good and is very welcome!

PR #314, opened 2021-04-02
Actix4 support

sfisol:


waled2020:
hi - small question: are paperclip became compatible with actix-web 4 beta version or not?
thanks in advance

dsferruzza:
> hi - small question: are paperclip became compatible with actix-web 4 beta version or not? if not: how long it takes to be compatible with actix web 4 beta?
> thanks in advance

@waled2020 paperclip's latest `master` (20ec2e954dfb15613807130105a41e8f4b91066e) is not compatible with actix-web 4; I guess that is the point of this PR (making it compatible with both actix-web 3 and 4).

---

@sfisol I am no maintainer but I would love to test your PR on my project. Would that be possible for you to rebase on `master` and use the latest actix-* betas?

iuslkae:
any dots? @wafflespeanut

sfisol:
@dsferruzza Sorry I was offline for more than week, I'll rebase in few hours.

sfisol:
I've updated to `actix 4.0.0-beta.8`
If someone need `actix-multipart` support, [there is branch actix4-multipart on my fork](https://github.com/sfisol/paperclip/tree/actix4-multipart) also rebased.

tomsik68:
Thank you so much for your effort sfisol I just tested with `actix-web 4.0.0-beta8` and it works flawlessly. I got a valid v2 spec which is readable by swagger editor, summary & description are loaded from rustdoc comments and appropriate response type is filled in the spec! :tada: 
Please let me know if there's anything I can help with!

simoneromano96:
Just a heads up, in the new actix-web the current `HttpResponse` has two different meanings: 
I believe the library should expose `actix_web::HttpResponse;` and not `actix_web::web::HttpResponse;` because the latest is actually the `actix-http` structure which works very differently from the current (3.x) `actix_web::HttpResponse` to mantain the same behaviour.

mharkins-cosm:
I pulled most of this into my forked copy (along with my own changes) and it worked for me.  Well done!  Would like to see this make it into a release so that I can stop running on my fork.

dsferruzza:
@sfisol It seems that `actix-web-4.0.0-beta.14` breaks compilation. An chance you could do another rebase of your changes?

FYI: your fork was working really well for me since `actix-web-4.0.0-beta.8` 👌 I hope it will be merged soon!

sfisol:
Hi @dsferruzza I've rebased on new paperclip and updated to actix-4.0.0-beta.14.

It seems to work.
Tests are not passing because of - as I think - some weird actix/actix4 features mangling which I did not spot yet.

Unfortunately I've checked it only on 1 project, because for others I'm struggling with actix-cors now - even if new version `0.6.0-beta.5` (compatible with actix-web beta 14) is now released, it gives me errors at runtime. I need to overcome this and then I'll prepare also actix-files4 and actix-multipart4 features.

@simoneromano96 - sorry I missed your comment. Paperclip re-exports `HttpResponse` from `web` module because it is a `web` module which gets "wrapped". So, if you want to use `actix_web::web::HttpResponse` then you can import it from paperclip's `web` module. If you want to use the other `HttpResponse` then just import `actix_web::HttpResponse`. If this introduces any problem please explain :-)

dsferruzza:
Hey @sfisol 

Sorry to be the one to ask you this again: `actix-web 4.0.0-beta.18` seems to break your branch, could you fix it, please?

It looks like a type parameter is no longer needed for the `Handler` trait:
image

Have a nice day!

sfisol:
Hi, @dsferruzza 

It turned out that new `Handler` works even better with paperclip :-) But I haven't time yet to test it. Give me a shout if you spot something wrong.

Version with additional actix modules here: [beta18-multipart-files-session](https://github.com/sfisol/paperclip/tree/actix4-beta18-multipart-files-session)

dsferruzza:
@sfisol Build breaks again, this time with `actix-web 4.0.0-beta.20`. It seems less trivial to solve, though:

image

sfisol:
Hi @dsferruzza,

On the contrary! This is just additional `+ 'static` bounds, which paperclip needs to follow.

Stay tuned for the re-base soon too :-)

sfisol:
Updated to actix-web 4.0.0-rc.3 and rabased.
Version with -multipart, -files, -session in `actix4-rc3-multipart-files-session` branch

cheesycod:
error[E0728]: `await` is only allowed inside `async` functions and blocks
  --> src/main.rs:88:27
   |
75 | #[api_v2_operation]
   | ------------------- this is not `async`

sfisol:
Hi @cheesycod , what kind of object do you return from this handler? I've tried to return different things in such defined handler and it works fine for me.

cheesycod:
The full code im trying to document with paperclip (i removed it after getting the errors so it would still compile):

```rust


#[doc(hidden)]
use actix_web::{get, web, http, HttpResponse, HttpRequest, error::ResponseError, Responder};
#[doc(hidden)]
use crate::models;
use std::collections::HashMap;

#[get("/index")]
/// Returns the index page
async fn index(req: HttpRequest, info: web::Query<models::IndexQuery>) -> impl Responder {
    let mut index = models::Index {
        top_voted: Vec::new(),
        certified: Vec::new(),
        new: Vec::new(),
        tags: Vec::new(),
        features: HashMap::new(),
    };

    let data: &models::AppState = req.app_data::<web::Data<models::AppState>>().unwrap();

    if info.target_type.as_ref().unwrap_or(&"bot".to_string()) == "bot" {
        index.top_voted = data.database.index_bots(models::State::Approved).await;
        index.certified = data.database.index_bots(models::State::Certified).await;
        index.tags = data.database.bot_list_tags().await;
        index.new = data.database.index_new_bots().await;
        ( 
            web::Json(index),
            http::StatusCode::OK,
        )
    } else {
        index.top_voted = data.database.index_servers(models::State::Approved).await;
        index.certified = data.database.index_servers(models::State::Certified).await;
        index.new = data.database.index_new_servers().await;
        index.tags = data.database.server_list_tags().await;
        (
            web::Json(index),
            http::StatusCode::OK,
        )
    }
}

#[get("/code/{vanity}")]
async fn get_vanity(req: HttpRequest, code: web::Path<String>) -> HttpResponse {
    let data: &models::AppState = req.app_data::<web::Data<models::AppState>>().unwrap();
    let resolved_vanity = data.database.resolve_vanity(&code.into_inner()).await;
    match resolved_vanity {
        Some(data) => {
            return HttpResponse::build(http::StatusCode::OK).json(data);
        }
        _ => {
            let error = models::CustomError::NotFoundGeneric;
            return error.error_response();
        }
    }
}

```

cheesycod:
> Make sure you're using the same `actix_web` version as `paperclip` uses. (For example check `Cargo.lock` if you have only `4.0.0-rc3` version there). Import `get` macro and `web` module from `paperclip::actix` rather than from `actix_web` directly. If the problem still persists create a repository with your example failing to compile, I'm curious what's going on :-)

I'm getting a new error now: error[E0277]: the trait bound `(actix_web::web::Json<models::Index>, StatusCode): TypedData` is not satisfied
  --> src/core.rs:14:1
   |
14 | #[api_v2_operation()]
   | ^^^^^^^^^^^^^^^^^^^^^ the trait `TypedData` is not implemented for `(actix_web::web::Json<models::Index>, StatusCode)`

cheesycod:
Repo is https://github.com/Fates-List/api-v3

sfisol:
Link doesn't work. Is the repo private?

cheesycod:
> Link doesn't work. Is the repo private?

oh sorry, its public now

sfisol:
I think it's because paperclip doesn't support handlers that return tuples. Change your handler to return bare `web::Json<models::Index>` and it should work.

cheesycod:
Ah, ok, can this be a feature request

chamoretto:
Since [actix 4.0 was released](https://github.com/actix/actix-web/blob/master/actix-web/CHANGES.md#400---2022-02-25), could this PR be updated and merged? Or there are some needed changes?

sfisol:
It's now ready for review

Cargo.toml
```diff
@@ -59,12 +62,14 @@ env_logger = "0.8"
 
 [features]
 # actix-web support
-actix = ["actix3"]
-actix-nightly = ["actix3-nightly"]
+actix = ["actix4"]
```
tiagolobocastro, line 65: I guess this means actix4 is a breaking change?

book/actix-plugin.md
```diff
@@ -163,15 +139,15 @@ curl http://localhost:8080/api/spec/v2
...
-      },
-      "parameters": [{
-        "in": "body",
-        "name": "body",
-        "required": true,
-        "schema": {
-          "$ref": "#/definitions/Pet"
-        }
-      }]
+        "parameters": [{
```
tiagolobocastro, line 142: nit: missing a comma?

core/src/v2/actix.rs
```diff
@@ -168,6 +173,40 @@ macro_rules! impl_empty({ $($ty:ty),+ } => {
     )+
 });
 
+#[cfg(feature = "actix4")]
+// Workaround for possibility to directly return HttpResponse from handler
+// after actix removed impl Future from HttpResponse.
+// https://github.com/actix/actix-web/pull/2601
+pub struct HttpResponseWrapper(pub HttpResponse);
```
tiagolobocastro, line 180: Not sure if this should be in paperclip itself, is this a requirement?

book/actix-plugin.md
```diff
@@ -9,48 +9,21 @@ Let's start with a simple actix-web application. It has `actix-web` and `serde`
...
-    ).bind("127.0.0.1:8080")?
-    .run().await
-}
-```
-
-Now, let's modify it to use the plugin!
-
-```rust
-use actix_web::{App, HttpServer};
+use actix_web::{App, HttpServer, Error};
```
tiagolobocastro, line 26: Any reason to remove the rest of the code? (sorry If I'm misinterpreting)

Cargo.toml
sfisol: Yeah, I think the same breaking change was introduced when changing from 2 to 3.

book/actix-plugin.md
sfisol: Yeah, sorry, I was manually changing keys order in JSON to not generate excessive diffs. I'll amend.

core/src/v2/actix.rs
sfisol: I can't figure out a way of not using it (it is used now in tests), and it's also a handy wrapper for end-user, so I inserted it here.

book/actix-plugin.md
sfisol: You're right, I somehow deleted the pre-actix example. I'll amend

core/src/v2/actix.rs
tiagolobocastro: On the linked PR @robjtede mentions the impl is useless in practice and removed it so just based on that my first stance would probably be for not adding it here if possible.
But if it is needed for us then I'm ok with it - I'll try to take a better look at this over the weekend.

robjtede: [See note in migration docs.](https://github.com/actix/actix-web/blob/master/actix-web/MIGRATION-4.0.md#returning-httpresponse-synchronously)

Is there a particular reason this is required?

sfisol:
Anybody knows why the build isn't working? It works for me locally

tiagolobocastro:
I think the nightly is broken also on master, I'll have a look tonight.

tiagolobocastro:
Wasn't expecting to see this:

`  = note: /usr/bin/ld: _ZN12actix_server6worker17MAX_CONNS_COUNTER7__getit5__KEY17hb572f826fbed2887E: TLS definition in /home/runner/work/paperclip/paperclip/target/debug/deps/libactix_server-56ba70876724c73e.rlib(actix_server-56ba70876724c73e.actix_server.ef5d9d00-cgu.0.rcgu.o) section .tbss._ZN12actix_server6worker17MAX_CONNS_COUNTER7__getit5__KEY17hb572f826fbed2887E mismatches non-TLS reference in /home/runner/work/paperclip/paperclip/target/debug/deps/libactix_server-56ba70876724c73e.rlib(actix_server-56ba70876724c73e.actix_server.ef5d9d00-cgu.8.rcgu.o)
          /usr/bin/ld: /home/runner/work/paperclip/paperclip/target/debug/deps/libactix_server-56ba70876724c73e.rlib: error adding symbols: bad value
          collect2: error: ld returned 1 exit status`

core/src/v2/actix.rs
sfisol: When I migrate inline handler to the:

```rust
web::to(|| async { HttpResponse::Ok().finish() })
```

(keep in mind that here `web` is a wrapper for original `web` from actix)

then I get 

```
    --> tests/test_app.rs:2503:36
     |
2503 |                         web::get().to(move || {
     |                                    ^^ the trait `Apiv2Operation` is not implemented for `impl futures::Future<Output = [async output]>`
```

I don't know how to impl `Apiv2Operation` for such type.

We have:

(1) `impl Future for ResponseWrapper<...>` and (2) `impl Apiv2Operation for ResponseWrapper<F, ...>`

`impl Apiv2Operation for F, where F::Future: Future<Output = HttpResponse<BoxBody>>` would probably resolve the problem but it will conflict with (2) because of (1).

When I introduced `HttpResponseWrapper` and implement `Responder` for it, it resolves the situation. I don't know why pure `HttpResponse` is not falling into (1) implementation even though it implements Responder too...

I also tried pin-boxing the result and `impl Apiv2Operation for Pin<Box<dyn Future<...>>`, but with no result.

tiagolobocastro:
I can repro the failure locally by using latest rust on our master branch:
`nix-shell --run 'cargo build --all --features "actix3 cli chrono url uuid swagger-ui v3"' --argstr date latest `
Might this be a compiler bug? maybe we should pin the nightly to an older version? What do you reckon @sfisol ?

sfisol:
Sure, but I think it should be done only temporarily

tiagolobocastro:
Agreed! Be nice if we had a bot to bump it weekly :smile:

Kavan72:
any news ? It would be nice to have it

core/src/v2/actix.rs
tiagolobocastro: any thoughts @wafflespeanut ?

tiagolobocastro: @sfisol should those comments above `HttpResponseWrapper` become doc comment?
I'm trying to think what's the best way to document this, maybe we should add a little example on the quick start?

sfisol: If you try to do it without wrapper then you get error

```
expected struct `HttpResponseWrapper`, found struct `HttpResponse
```

so user is directed to the wrapper by compiler, so I just added a minimal docstring example, and I think this will be sufficient. 

I can't figure out though how to write docstring example that will be tested - `actix_web` and `paperclip` crates are not available in docstring, I don't know why.

Cargo.toml
tiagolobocastro: I think we should change this going forward, otherwise we'll keep having breaking changes.

Also, we have a problem for testing, example one of the tests requires actix:
```
required-features = ["cli", "actix", "uuid", "chrono", "swagger-ui"]
```
This means we don't run the tests for previous versions, which seems wrong, and explains why we had a few actix2 bugs that should have been caught.

I propose that the actix version should be explicitly specified in the feature: "actix3" or "actix4".
Then we can have an "actix" feature to allow tests to run for all actix versions - which could just be the "actix-base" we already have:
```
required-features = ["cli", "actix-base", "uuid", "chrono", "swagger-ui"]
```

piotr-d (approved):
Looks great, approved.

Cargo.toml
sfisol: We do test previous versions now. I've added it in this PR.
Of course I can remove bare 'actix' feature, but this will be breaking change too. The last one though :-) 
PS. Changing 'actix' feature to take a role of 'actix-base' will be a "silent braking change" so I'm against it.
What do you think?

tiagolobocastro: We do test previous but not for test_app, because "actix" is a required feature for this test to run, which means test_app tests don't run for previous versions?

sfisol: Ok, I didnt understand you in the first place. So, things to be done:

1. Remove 'actix' alias-feature
2. Do tests for actix-base

Confirm and I'll amend.

tiagolobocastro: Sorry about that, eloquence was never my strong suite :D 
Yep I'm happy with that!

plugins/actix-web/Cargo.toml
```diff
@@ -13,19 +13,22 @@ repository = "https://github.com/paperclip-rs/paperclip"
...
+actix-web4 = { version = "4", default-features = false, optional = true, package = "actix-web" }
 parking_lot = ">=0.10, <0.13"
 serde_json = "1.0"
 once_cell = "1.4"
 openapiv3 = { version = "1.0.0-beta.5", optional = true }
 include_dir = { version = "0.7.2", optional = true }
 
 [features]
-actix = ["actix3"]
+actix = ["actix4"]
```
tiagolobocastro, line 28: Should we also remove "actix" for these other libs? Alternatively we could leave actix pointing to actix3 and then we wouldn't break compat?

tiagolobocastro (approved):
Thanks @sfisol :partying_face:

plugins/actix-web/Cargo.toml
sfisol: Yes, I just removed unused `actix` features. The place where it lasts are macros, but it is used internally so I think it's OK. I also updated `Makefile` and `actix-plugin.md` to use `actix4` feature.

fritterhoff:
🥳

Chaostheorie:
Thank you, @sfisol, for implementing the update. Could you please publish a new release for paperclip with the support for actix-web 4 @tiagolobocastro?

tiagolobocastro:
@Chaostheorie will publish when #404 is addressed.

PR #331, opened 2021-09-02
swagger-ui: Add feature 

patrickelectric:
Inspired by #228


patrickelectric:
ping @tiagolobocastro

patrickelectric:
ping @wafflespeanut

patrickelectric:
ping @tiagolobocastro @wafflespeanut

patrickelectric:
@tiagolobocastro rebased

plugins/actix-web/src/lib.rs
```diff
@@ -281,6 +293,42 @@ where
...
+    ///
+    /// **NOTE:** you **MUST** call with_json_spec_at before calling this function
+    #[cfg(feature = "swagger-ui")]
+    pub fn with_swagger_ui_at(mut self, path: &str) -> Self {
+        let spec_path = self.spec_path.clone().expect(
+            "Specification not set, be sure to call `with_json_spec_at` before this function",
+        );
+
+        let path: String = path.into();
+        let regex_path = format!("{}/{{filename:.*}}", path);
```
tiagolobocastro, line 306: what is filename meant to be here?

patrickelectric: Hi @tiagolobocastro I have updated with a comment:

// Grab any file request from the documentation UI path and fetch it from SWAGGER_DIST
// E.g: js, html, svg and etc.

So when you have /swagger_ui/index.html or /swagger_ui/icons.svg, we can use the filename to fetch from the SWAGGER_DIST

tiagolobocastro: So what are we meant to pass to `with_swagger_ui_at` ? I would expect that we simply provide a path where we want the swagger_ui to be accessible, ex: /v0/swagger, and in that case it seems that we get redirected to the index.htlm?

So why would we need to specify a particular file?

tiagolobocastro:
Btw, could you please add a test, ex in tests/test_app.rs? Doesn't have to be very complicated, just to check that we can get to the url. And also please update the changelog, this is very worth mentioning there :)

plugins/actix-web/src/lib.rs
patrickelectric: So, what happens is the following, `/v0/swagger` will be redirect to `/v0/swagger/index.html` with the url query parameter, the query will be populated with the path of the openapi/swagger definition file from `with_json_spec_at`. The redirect is also necessary to have a filename to fetch the file from SWAGGER_DIST and have a valid regex match.

patrickelectric:
> Btw, could you please add a test, ex in tests/test_app.rs? Doesn't have to be very complicated, just to check that we can get to the url. And also please update the changelog, this is very worth mentioning there :)

Will do!

plugins/actix-web/src/lib.rs
tiagolobocastro: Oh right, got it :+1:

patrickelectric:
@tiagolobocastro done!

tiagolobocastro (approved):
:rocket:

Issue #204, opened 2020-07-21
Generic types in generated spec show only the definition used in the last configured route


alexabstreiter:
Hey there, thanks for this great project!

I am using paperclip to generate the OpenAPI spec for my rust backend.
In order to have consistent API responses, I am using a generic `DataResponse<T>` struct. Unfortunately, in the API spec generated by paperclip, the type T is always linked to the type that was used in the last configured route.

Minimal reproducible example:
```
#[derive(Serialize, Deserialize, Debug, Apiv2Schema)]
struct DataResponse<T> {
    pub data: T,
}

#[derive(Serialize, Deserialize, Debug, Apiv2Schema)]
struct ResourceResponse {
    pub name: String,
}

#[derive(Serialize, Deserialize, Debug, Apiv2Schema)]
struct CreateResourceResponse {
    pub id: Uuid,
}

#[api_v2_operation]
async fn create_resource() -> Result<web::Json<DataResponse<CreateResourceResponse>>, ApiError> {
    Ok(web::Json(DataResponse {
        data: CreateResourceResponse { id: Uuid::new_v4() },
    }))
}

#[api_v2_operation]
async fn get_resource(
    _id: web::Path<Uuid>,
) -> Result<web::Json<DataResponse<ResourceResponse>>, ApiError> {
    Ok(web::Json(DataResponse {
        data: ResourceResponse {
            name: "Test".to_owned(),
        },
    }))
}

pub fn routes(cfg: &mut web::ServiceConfig) {
    cfg.service(
        web::scope("/api").service(
            web::scope("/resource")
                .service(
                    web::scope("/{id}")
                        .service(web::resource("").route(web::get().to(get_resource))),
                )
                .service(web::resource("").route(web::post().to(create_resource))),
        ),
    );
}

#[actix_rt::main]
async fn main() -> std::io::Result<()> {
    /* ... */
    HttpServer::new(move || {
        App::new()
            .wrap_api()
            .with_json_spec_at(SWAGGER_ENDPOINT)
            .configure(routes)
            .build()
    })
    .bind(binding_address)?
    .run()
    .await
}
```

The generated API spec looks like this:
```
swagger: '2.0'
definitions:
  DataResponse:
    properties:
      data:
        properties:
          id:
            type: string
            format: uuid
        required:
          - id
    required:
      - data
paths:
  /api/resource:
    post:
      responses:
        '200':
          description: OK
          schema:
            $ref: '#/definitions/DataResponse'
[...]

PR #332, opened 2021-09-06
fix(generics): support generic response types

mharkins-cosm:
This is an attempt to address issue #204.  Because the nested generic objects must be checked dynamically for names, I had to get rid of the `NAME` constant and replace it with a `name()` method.  Then it's just including named type parameter names in the parent name surrounded by angle brackets (`<>`).


tiagolobocastro:
@mharkins-cosm could you please rebase so we can run the tests with gha?
I think this would be a breaking change to the API, so we'd have to bump up the version and add this to the changelog.

mharkins-cosm:
> > > @mharkins-cosm could you please rebase so we can run the tests with gha? I think this would be a breaking change to the API, so we'd have to bump up the version and add this to the changelog.
> > 
> > 
> > Done
> 
> Cool, it's running now. Could you please also add tests for generic response types and also update the changelog? Thanks

Yes, I will.  I'm sorry I haven't had time to get to it yet.  It's on my todo list!

mharkins-cosm:
@tiagolobocastro Finally got to adding the requested test and changelog.  Would you mind having a look and kicking of the checks if it looks good to you?

mharkins-cosm:
saw the lint check failed.  i think i should have that fixed now.

macros/src/actix.rs
```diff
@@ -748,7 +748,20 @@ pub fn emit_v2_definition(input: TokenStream) -> TokenStream {
...
+    let base_name = extract_rename(&item_ast.attrs).unwrap_or_else(|| name.to_string());
+    let type_params: Vec<&Ident> = generics.type_params().map(|p| &p.ident).collect();
+    let schema_name = if type_params.is_empty() {
+        quote! { #base_name }
+    } else {
+        let type_names = quote! {
+            vec![#(#type_params::name()),*]
+                .iter()
+                .filter(|n| n.is_some())
+                .map(|n| n.as_ref().unwrap().clone()).collect::<Vec<String>>()
```
tiagolobocastro, line 760: could use filter_map here? also maybe into_iter instead of iter avoiding having to write the clone?

tiagolobocastro:
> saw the lint check failed. i think i should have that fixed now.

cool I kicked the GHA, will have a another look tomorrow

macros/src/actix.rs
mharkins-cosm: hmmm... i get a warning for `into_iter`, but can use `filter_map`.

mharkins-cosm: ✅

mharkins-cosm:
Funny the clippy warning is not part of this PR, nor does it appear to be accurate.  Do you want me to add an `allow` annotation anyway, to allow the check to pass?

tiagolobocastro:
> Funny the clippy warning is not part of this PR, nor does it appear to be accurate. Do you want me to add an `allow` annotation anyway, to allow the check to pass?

That'd be great if you don't mind 👍

mharkins-cosm:
Thanks very much @tiagolobocastro ! Any idea when it'll be released, along with the actix 4 support?

tiagolobocastro:
No, thank you! I'll probably try to publish it this weekend.

PR #344, opened 2021-10-04
properties with a leading '@' should be renamed to 'at_' to allow both variants (e.g. @type and type) at the same time

### LINKED ISSUE #343: properties with @ (e.g. @type) should be distinguishable from those without @
The APIs from https://www.tmforum.org/ make use of properties with a leading @, e.g. `@type`. 

At the same time, the same objects might have a property `type`. The one with the leading @ should be named `at_type' in the produced rust struct.

jens1205:
fixes #343

### DIFF
--- a/src/v2/codegen/object.rs
+++ b/src/v2/codegen/object.rs
@@ -24,6 +24,9 @@ use std::{
 /// Regex for appropriate escaping in docs.
 static DOC_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r"\[|\]").expect("invalid doc regex?"));
 
+/// Regex for renaming properties with leading @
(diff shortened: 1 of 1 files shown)


src/v2/codegen/object.rs
```diff
@@ -233,6 +236,16 @@ pub struct ObjectField {
     pub child_req_fields: Vec<String>,
 }
 
+pub fn to_snek_case(name: &str) -> String {
```
tiagolobocastro, line 239: did you mean snake_case?

jens1205: for some reason the old code also used to_snek_case from https://docs.rs/heck/0.3.0/heck/trait.SnekCase.html , so I named it also like this. But after looking into this crate, we could switch to use to_snake_case everywhere, as snek_case is only an alias.

jens1205: I changed all `snek_case` to `snake_case`

Issue #357, opened 2021-11-11
[Actix] Type definition overriden when struct used with same name multiple times


TobiasDeBruijn:
Hey there,

Thanks for all the hard work on Paperclip. I'm using it together with Actix to generate a spec for my API, however I've noticed something odd. Almost every request handler (split out into modules) has a `Request` and `Response` struct, for the request and response respectively. However it seems to be only provided once in the final spec. 

Assume I have the following two endpoints:
```rs
// foo.rs
#[derive(Deserialize, Apiv2Schema)]
pub struct Request {
    a: i32,
    b: i32
}
...

// bar.rs
#[derive(Deserialize, Apiv2Schema)]
pub struct Request {
    c: String,
    d: f64
}
```
Both API will show the same request struct, rather than foo having it's own, and bar having it's own

## Potential solution
I think a potential solution would be to include the module in the definition name, that way you can never have overlap

## Versions
actix: 3.3.2
paperclip: 0.6.1
rustc: 1.57.0-beta

PR #358, opened 2021-11-11
[Actix plugin] Add random prefix to definition in API spec 

TobiasDeBruijn:
This PR adds a 6 character (A-Z) prefix to every generated definition (the `Apiv2Schema` proc macro). The reason for this PR is described in #357, this PR solves #357 too.

Feedback is highly appreciated!

### DIFF
--- a/macros/src/actix.rs
+++ b/macros/src/actix.rs
@@ -6,6 +6,7 @@ use http::StatusCode;
 use lazy_static::lazy_static;
 use proc_macro::TokenStream;
 use quote::{quote, ToTokens};
+use rand::Rng;
 use strum_macros::EnumString;
 use syn::{
     parse_macro_input,
@@ -15,7 +16,6 @@ use syn::{
     Generics, Ident, ItemFn, Lit, Meta, MetaList, MetaNameValue, NestedMeta, Path, PathArguments,
     ReturnType, Token, TraitBound, Type, TypeTraitObject,
 };
-use rand::Rng;
 
 use proc_macro2::TokenStream as TokenStream2;
 use std::collections::HashMap;
--- a/macros/src/actix.rs
+++ b/macros/src/actix.rs
@@ -15,6 +15,7 @@ use syn::{
     Generics, Ident, ItemFn, Lit, Meta, MetaList, MetaNameValue, NestedMeta, Path, PathArguments,
     ReturnType, Token, TraitBound, Type, TypeTraitObject,
 };
+use rand::Rng;
 
 use proc_macro2::TokenStream as TokenStream2;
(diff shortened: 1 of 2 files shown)


macros/src/actix.rs
```diff
@@ -719,10 +720,13 @@ pub fn emit_v2_definition(input: TokenStream) -> TokenStream {
     };
 
     let schema_name = name.to_string();
+    let random_prefix = generate_random_uppercase(6);
```
tiagolobocastro, line 723: not sure this is the right thing to do here, otherwise we're generating different specs on different builds?

TheDutchMC: That is a good point. That is indeed the case here. For what I've read and heard, it's not possible to get the fully qualified path in a proc macro.

Do you have any suggestions for another solution?

tiagolobocastro: Even a module name is probably not ideal as we want the type name to be named somewhat appropriately, since it's part of an API - what about a way to explicitly rename the openApi type? eg:
```rust
#[derive(Deserialize, Serialize, Apiv2Schema)]
#[apiv2(rename = "PetThatSings")]
struct Pet { }
```
It would also be helpful if we could issue a warning or error if we encounter a duplicate when stitching up the spec.

TheDutchMC: That could also definitely be a good option, though I don't know how to implement this myself. Could we maybe 'compute' a hash of sorts and use that as prefix? That way the name is consistent and only changes when the struct itself does.

This means a user isn't required to rename every struct, which gets tedious, but it doesnt leave everything random either.

TheDutchMC: I have figured out a way to get the fully qualified path, so that is now too an option

tiagolobocastro: How would the type names look like? 
That would be better than random, but that's still something we wouldn't want to do by default. Maybe we could have an attribute enables using that fqp?
```rust
#[derive(Deserialize, Serialize, Apiv2Schema)]
#[apiv2(prefix_fqp)]
struct Pet { }
or 
#[derive(Deserialize, Serialize, FQPApiv2Schema)]
struct Pet { }
```

TheDutchMC: I haven't generated it out yet, but by default we get e.g foo::bar::Request, but we could emit those double colons (it probably does not work well with yaml, so we should I think), or replace them with e.g hyphens.

An attribute that enables it would be the best solution I think, additionally we could stick it behind a feature flag too.

tiagolobocastro:
Replaced by https://github.com/paperclip-rs/paperclip/pull/373

PR #364, opened 2021-11-23
updates for openapiv3 1.0.0

ahl:
I'm getting the https://github.com/glademiller/openapiv3 crate ready for a 1.0.0 release. To that end, I wanted to make sure consumers were ok with the changes. As one of the more popular consumers, would you be amenable to taking a look? If you're happy enough with it, I'll do some final checks and publish 1.0.0 in the next couple of weeks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ url_dep = { version = ">=1.7,<3", package = "url" }
 thiserror = "1.0"
 anyhow = "1.0"
 once_cell = "1.4"
-openapiv3 = { version = "0.4.0", optional = true }
+openapiv3 = { version = "1.0.0-beta.5", optional = true }
 indexmap = { version = "1.0", features = ["serde-1", "std"], optional = true }
 
 [dev-dependencies]
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -32,7 +32,7 @@ url = { version = "2", optional = true }
 uuid = { version = "0", optional = true }
 thiserror = "1.0"
 serde_qs = { version = "0", optional = true }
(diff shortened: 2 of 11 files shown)


tiagolobocastro:
@ahl that looks good thanks.
Btw, very interesting the change for the security to be an Option, I found the related PR :+1: - for futures updates of the openapi3 I wonder if it'd possible to have some kind of gh action/script that would parse changes to the openapi3 spec.md and raise a PR to at least the doc comments updated or something like that.

Based on some "miscreated PR's" I think @nirhaas might be using the openapiv3 feature, so maybe @nirhaas is willing to take a look as well?

ahl:
Cool; thanks

nirhaas:
I don't think I have much to offer you guys in terms of code review. I liked the Option for security as well. 
Not sure if it's in the scope of this PR, but it seems like `impl From<v2::DefaultApiRaw> for openapiv3::OpenAPI` is not handling the `security` keyword in the root level (i.e. apply security to all operations). I would expect to see something like this on openapi.rs:
```rust
spec.security = if v2.security.is_empty() {
                None
            } else {
                Some(
                    v2.security
                        .iter()
                        .map(|s| {
                            s.iter().fold(indexmap::IndexMap::new(), |mut i, (k, v)| {
                                i.insert(k.to_string(), v.clone());
                                i
                            })
                        })
                        .collect(),
                )
            };
```

ahl:
I've updated this to 1.0.0 if you want to consider merging it.

tiagolobocastro:
thanks @ahl, I've just kicked the CI and will merge if it passes - it seems our clippy nightly might be broken so I'll try to fix that first.

PR #367, opened 2021-12-08
add #[openapi(rename = "...")] attribute

tomsik68:
Hello! 

I designed a fix for #357 which is a little different from the approach in #358. This time, I'm adding the explicit rename using `#[openapi(rename = "...")]`. If there's a good way to test my approach using existing paperclip tests, I'd love to know about it.

### DIFF
--- a/macros/src/actix.rs
+++ b/macros/src/actix.rs
@@ -661,34 +661,19 @@ pub fn emit_v2_errors_overlay(attrs: TokenStream, input: TokenStream) -> TokenSt
 
 fn extract_rename(attrs: &[Attribute]) -> Option<String> {
     let attrs = extract_openapi_attrs(attrs);
-    for punc in attrs {
-        for attr in punc {
-            if let NestedMeta::Meta(attr) = attr {
-                match attr {
-                    Meta::NameValue(nv) => {
-                        if nv
-                            .path
-                            .get_ident()
-                            .map(|id| id.to_string() == "rename")
-                            .unwrap_or(false)
-                        {
-                            match nv.lit {
-                                Lit::Str(s) => {
-                                    return Some(s.value());
-                                }
-                                _ => {
-                                    emit_error!(
-                                        nv.lit.span().unwrap(),
-                                        format!(
-                                            "`#[{}(rename = \"...\")]` expects a string argument",
-                                            SCHEMA_MACRO_ATTR
-                                        ),
-                                    );
-                                }
-                            }
-                        }
-                    }
-                    _ => {}
+    for attr in attrs.flat_map(|attr| attr.into_iter()) {
+        if let NestedMeta::Meta(Meta::NameValue(nv)) = attr {
+            if nv.path.is_ident("rename") {
+                if let Lit::Str(s) = nv.lit {
+                    return Some(s.value());
+                } else {
+                    emit_error!(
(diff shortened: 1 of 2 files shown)


macros/src/actix.rs
```diff
@@ -659,6 +659,44 @@ pub fn emit_v2_errors_overlay(attrs: TokenStream, input: TokenStream) -> TokenSt
     gen.into()
 }
 
+fn extract_rename(attrs: &[Attribute]) -> Option<String> {
```
tiagolobocastro, line 662: We already have an `extract_openapi_attrs` - could we make use of it here?
Btw, while you're at it could you replace the string "openapi" in `extract_openapi_attrs` with `SCHEMA_MACRO_ATTR`?

macros/src/actix.rs
```diff
@@ -659,6 +659,44 @@ pub fn emit_v2_errors_overlay(attrs: TokenStream, input: TokenStream) -> TokenSt
...
+                Meta::NameValue(mnv) => {
+                    if mnv
+                        .path
+                        .get_ident()
+                        .map(|id| id == "rename")
+                        .unwrap_or(false)
+                    {
+                        match mnv.lit {
+                            Lit::Str(s) => Some(s.value().clone()),
+                            _ => panic!("#[openapi(rename = ...)] expects literal as argument"),
```
tiagolobocastro, line 686: Should probably use `emit_error` instead of panic

macros/src/actix.rs
```diff
@@ -659,6 +659,44 @@ pub fn emit_v2_errors_overlay(attrs: TokenStream, input: TokenStream) -> TokenSt
...
+fn extract_rename(attrs: &[Attribute]) -> Option<String> {
+    attrs
+        .into_iter()
+        .filter(|attr| {
+            attr.path
+                .get_ident()
+                .map(|id| id.to_string() == "openapi")
+                .unwrap_or(false)
+        })
+        .next()
```
tiagolobocastro, line 671: This doesn't seem right, we should probably search the it as it might not always the first one? (we could add other things in the future).
Maybe have a look at `emit_v2_security` which uses the openapi attribute already.

tiagolobocastro:
> Hello!
> 
> I designed a fix for #357 which is a little different from the approach in #358. This time, I'm adding the explicit rename using `#[openapi(rename = "...")]`. If there's a good way to test my approach using existing paperclip tests, I'd love to know about it.

So the tests basically take your "config", generate the openapi spec and compare it against a hardcoded spec using `check_json`. I see you've renamed a Pet - that should fail the tests now are they expect to see `Pet` and not `PetRenamed` - however please add a new test instead to avoid changing all the existing tests which expect `Pet`.

tomsik68:
Hey @tiagolobocastro, I rebased, reimplemented `extract_rename` using your suggestions and added a test case for it! Could you please take another look when you have a chance? Thanks!

tiagolobocastro:
awesome thanks @tomsik68, I've kicked gh actions and will try to have a look during the week

tiagolobocastro:
hey @tomsik68 would you mind fixing the clippy errors [here](https://github.com/paperclip-rs/paperclip/runs/4929089935?check_suite_focus=true)

tiagolobocastro:
Thank you @tomsik68 !

PR #369, opened 2021-12-10
Fix favicons for swagger-ui and add simple error handling

LelouBil:
I noticed there was a lot of panics in my console when accessing the swagger-ui page and I tracked it down to the contents_utf8() call in the actix-web plugin.

I replaced it with a single contents() so it should work for any file type and not just text.
I also added an expect call with a message instead of just using unwrap()


tiagolobocastro:
Thanks @LelouBil, could you please fix the linter failure?

LelouBil:
It should be good now.

tiagolobocastro (approved):

PR #373, opened 2022-01-04
Add the module path to the definition name to avoid name collisions

### LINKED ISSUE #357: [Actix] Type definition overriden when struct used with same name multiple times

TobiasDeBruijn:
This PR intends to fix #357, but uses an alternative approach to #358.

With this PR, the behaviour will be as follows:
```rs
mod foo {
    pub mod bar {
        use paperclip::actix::{web, api_v2_operation, Apiv2Schema};
        use serde::Serialize;

        #[derive(Serialize, Apiv2Schema)]
        pub struct Response {
            bar: i32
        }

        #[api_v2_operation]
        pub fn b() -> web::Json<Response> {
            web::Json(Response { bar: 10 })
        }
    }
}
```
results in the following spec:
```yaml
  foo_bar_Response:
    type: object
    properties:
      bar:
        type: integer
        format: int32
    required:
      - bar
...
          schema:
            $ref: '#/definitions/foo_bar_Response' 
```
The advantage of this PR over #358 is that the generated spec is always the same for the same input code.

The name is determined by taking the output of `std::module_path!()`, removing the first 'component' (i.e. the crate name), then joining with `_` as a separator.

This **will not** break code backwards compatibility, but it **will** break spec backwards compatibility, i.e. the generated spec is different from how it would be without this PR


macros/src/actix.rs
```diff
@@ -720,26 +721,42 @@ pub fn emit_v2_definition(input: TokenStream) -> TokenStream {
...
+                    .filter(|(index, _)| *index != 0) // Skip the first element, i.e the crate name
+                    .map(|(_, component)| component)
+                    .collect::<Vec<_>>()
+                    .join("_");
+                format!("{}_{}", trimmed_module_path, #schema_name)
+            }
+        }
 
+        impl #impl_generics paperclip::v2::schema::Apiv2Schema for #name #ty_generics #where_clause {
+            //const NAME: Option<&'static str> = None;
```
tiagolobocastro, line 742: can remove this line?

tiagolobocastro (commented):
Hey @TobiasDeBruijn, as discussed previously this should be opt-in only - we can probably do that by hiding this behind a feature gate?

tiagolobocastro:
Looks good, can we add a test for this feature?

TobiasDeBruijn:
I'll give writing a test a shot!

PR #388, opened 2022-02-10
fix: add extensions to root level of spec

wzzrd:
as a follow up to #387 which added extensions to the Info object, this PR adds extensions to the root of the OpenAPI document.


tiagolobocastro:
thanks @wzzrd !

PR #389, opened 2022-02-12
Implement actix schema generator for tagged enums

### LINKED ISSUE #97: Support enum variants with named fields
As discussed [here](https://github.com/wafflespeanut/paperclip/issues/69#issuecomment-521974040): oneOf, anyOf, allOf.

Filing this issue in order to track when this will be supported.

Siphalor:
Resolves #97.

## About 
This pr implements schema generation for complex enums with serde's tag system.

Supported are all types of tagging ([see here](https://serde.rs/enum-representations.html)), except for the enum tuples.
Tuple schema generation is currently not implemented correctly anyway, so that should go into another pr.

This pr currently uses `anyOf` instead of `oneOf` to cover some possible edge cases, where serde would just use the first variant that matches (especially visible in `untagged` enums).

To be able to implement this, this pr adds `any_of` and `const` to the internal schema representation.

## Up for discussion
- Usage of `anyOf` vs. `oneOf` as described above.
- Currently the old enum system is kept for enums that use the default tagging (`external`) and that only have unit variants.
  This makes it effectively impossible to add descriptions for the enum variants.
  Possible workarounds:
  - Check for documentation on each variant and if present, go for the new `any_of` system.
  - Get rid of the old `enum` system and just use `any_of` for everything.

## Other notes
- The files that I worked on in this pr were incredibly huge, making editing more painful than it needs to be.
- Let me know if there's any more documentation needed :)

### DIFF
--- a/core/src/v2/schema.rs
+++ b/core/src/v2/schema.rs
@@ -54,6 +54,16 @@ pub trait Schema: Sized {
     /// - `serde_json::Value` works for both JSON and YAML.
     fn enum_variants(&self) -> Option<&[serde_json::Value]>;
 
+    /// More complex enum variants, where each has it's own schema.
+    fn any_of(&self) -> Option<&Vec<Resolvable<Self>>>;
+
+    /// A constant value for this schema. It's `serde_json::Value`
+    /// because:
+    ///
+    /// - Constants are allowed to have any value.
+    /// - `serde_json::Value` works for both JSON and YAML.
+    fn const_value(&self) -> Option<&serde_json::Value>;
+
     /// Returns whether this definition "is" or "has" `Any` type.
     fn contains_any(&self) -> bool {
         _schema_contains_any(self, vec![])
--- a/book/actix-plugin.md
+++ b/book/actix-plugin.md
@@ -185,7 +185,8 @@ Similarly, if we were to use other extractors like `web::Query<T>`, `web::Form<T
 
 #### Known limitations
 
-- **Enums:** OpenAPI (v2) itself supports using simple enums (i.e., with unit variants), but Rust and serde has support for variants with fields and tuples. I still haven't looked deep enough either to say whether this can/cannot be done in OpenAPI or find an elegant way to represent this in OpenAPI.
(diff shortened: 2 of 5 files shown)


tiagolobocastro:
Hey @Siphalor, please correct me if I'm wrong, but anyOf is not part of openapi version 2?

Siphalor:
Sadly you seem to be right.

It seems like I was over-enthusiastic and got lost in the JSON Schema spec without noticing that OpenAPIv2 doesn't support the latest draft.

Well, maybe this is at least useful as a reference for the v3 implementation then :)

tiagolobocastro:
Yes this is not all for naught, it is very useful indeed because we want to eventually output proper openapi v3. 
I added the initial dumb v2-v3 conversion but sadly for lack of time did not get to the next step.

core/src/v2/schema.rs
```diff
@@ -54,6 +54,16 @@ pub trait Schema: Sized {
     /// - `serde_json::Value` works for both JSON and YAML.
     fn enum_variants(&self) -> Option<&[serde_json::Value]>;
 
+    /// More complex enum variants, where each has it's own schema.
+    fn any_of(&self) -> Option<&Vec<Resolvable<Self>>>;
```
tiagolobocastro, line 58: rather than just any_of I wonder if we could have a schema kind that would allow for all the other variants (oneOf, etc)

tiagolobocastro: example:
```rust
pub enum SchemaKind {
    Type(crate::v2::models::DataType),
    OneOf {
        #[serde(rename = "oneOf")]
        one_of: Vec<crate::v2::models::Resolvable<Self>>,
    },
    AllOf {
        #[serde(rename = "allOf")]
        all_of: Vec<crate::v2::models::Resolvable<Self>>,
    },
    AnyOf {
        #[serde(rename = "anyOf")]
        any_of: Vec<crate::v2::models::Resolvable<Self>>,
    },
}
```

tiagolobocastro:
If you were so inclined, how about adding a new ApiV3Schema trait (eg in core/v3/schema.rs) containing these new additions?

PR #402, opened 2022-03-16
fix: Append scope services instead of replacing

### LINKED ISSUE #380: [Actix] If both a GET and POST is defined on the same path,  only one gets included
Noticed this issue today while checking the spec manually. I've got the following routes:
```rs
.route("[..]/wifi", web::get().to(endpoints::v1::[..]::get))
.route("[..]/wifi", web::post().to(endpoints::v1::[..]::set))
```
Where `[..]` indicates a path that has been left out. Only the `POST` endpoint is in the spec, the `GET` is not.

## Versions
- Actix: `3.3.3`
- Paperclip: Local fork (see #373), essentially v0.6.1
- Rust: `1.58`

fritterhoff:
So instead of inserting an item and overwriting the existing I simply append the existing maps.

Should fix #380

### DIFF
--- a/plugins/actix-web/src/web.rs
+++ b/plugins/actix-web/src/web.rs
@@ -457,7 +457,13 @@ where
                 op.set_parameter_names_from_path_template(&p);
             }
 
-            self.path_map.insert(p.clone(), map);
+            if let Some(mut existing) = self.path_map.get_mut(&p){
+                existing.methods.append(&mut map.methods);
+                existing.parameters.append(&mut map.parameters);
+            } else {
+                self.path_map.insert(p.clone(), map);
+            }
+
         }
 
         SecurityScheme::append_map(factory.security_definitions(), &mut self.security);
--- a/plugins/actix-web/src/web.rs
+++ b/plugins/actix-web/src/web.rs
@@ -457,13 +457,12 @@ where
                 op.set_parameter_names_from_path_template(&p);
             }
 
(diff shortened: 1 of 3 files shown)


TobiasDeBruijn (approved):

tiagolobocastro:
Thanks @fritterhoff! Could you please fix the " Lint / rustfmt (stable) (pull_request) Failing after 8s — rustfmt (stable) " ?

fritterhoff:
@tiagolobocastro done.

fritterhoff:
I will check the failing tests in a minute.

fritterhoff:
Actually I have no idea why that changed but I think that is not affected by my changes...?

tiagolobocastro:
> Actually I have no idea why that changed but I think that is not affected by my changes...?

The stable build and test should pass; the nightly has been failing for some time now so I think we can safely ignore it.

fritterhoff:
> > Actually I have no idea why that changed but I think that is not affected by my changes...?
> 
> The stable build and test should pass; the nightly has been failing for some time now so I think we can safely ignore it.

Yeah. I added a missing `/`. I thought was working some days ago 😕

fritterhoff:
So from my point of view we could merge this. I have no further ideas why the nightly builds fail.

tiagolobocastro:
Thanks @fritterhoff !

PR #406, opened 2022-03-24
fix: use only required actix-service

tiagolobocastro:


tiagolobocastro:
@Jawbone999

Jawbone999 (approved):
Looks like this should cut down on some cruft, which is always good! 👍 Thanks

PR #408, opened 2022-03-30
Fixed #407

### LINKED ISSUE #407: when declare routes with out prefix "/", the swagger doc doesn't add it automated
this is my code here, and the actix allows this gramma, though it recommends to use `#[post("/register")]`  with the prefix  `/`

```rust
#[api_v2_operation]
#[post("register")]
async fn register() -> HttpResponse {
    HttpResponse::Ok().body("test app")
}
```

and I think it's better for paperclip to add the prefix `/` automated if it doesn't exist.

image

TobiasDeBruijn:
## Summary
Solves #407 

When using proc macros to declare the path to a route, e.g:
```rs
#[post("foo")]
async fn foo() {}
```
No `/` is prefixed to the path `foo`. It is recommended by the Actix documentation to include the `/` in the path (i.e. `/foo`), but it is not a requirment and things will work just fine without it. However, to my knowledge it does need to be present in the OpenAPI documentation. This PR addresses this issue.

## The way it works
I've altered the proc macro which is invoked here. It now checks if the provided path starts with a `/`, if not it will add it.

## Backwards compatibility
Backwards compatibility will not be broken. Specs might change, if users were using 'unprefixed' paths, those paths will now be prefixed

## Tests
I've added a test in `tests/test_app.rs`: `test_simple_app`, this passes.

Thanks!

### DIFF
--- a/macros/src/actix.rs
+++ b/macros/src/actix.rs
@@ -1569,12 +1569,13 @@ impl super::Method {
         let variant: proc_macro2::TokenStream = self.variant().parse()?;
         let handler_name_str = handler_name.to_string();
 
-        let uri = uri.to_string()
-            .replace("\"", ""); // The uri is a string lit, which contains quotes, remove them
+        let uri = uri.to_string().replace('\"', ""); // The uri is a string lit, which contains quotes, remove them
 
-        let uri_fmt = if !uri.starts_with("/") {
+        let uri_fmt = if !uri.starts_with('/') {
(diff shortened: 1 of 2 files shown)


tiagolobocastro:
Looks good @TobiasDeBruijn, please address the clippy and lint failures before we merge.

TobiasDeBruijn:
DIdn't notice those, will do!

PR #413, opened 2022-04-07
Feat: basic support for header parameter

rlebran:
Hi,

Using paperclip I currently have an issue representing one of my API requirement. I would like to be able to document the fact that every request required some Headers.

This PR aims to implement a basic Header parameter support using the same model as Security definitions.

# How ?

To do so I did introduce a new derive macro, `Apiv2Header` which accepte attribute parameter through `#[openapi]`.
Supported parameters are currently "name", "description" and "format".

# Why is this a partial / basic support ?

Only String type is currently supported. Even if all data format are supported to describe the type, the associated logic is not (array handling with the different format for example).

# Example

An example can be found in the newly added test.

Here's an example on how to use the new derive macro:

```rust
#[derive(Apiv2Header, Deserialize)]
#[openapi(
    name = "X-Request-ID",
    description = "Allow to track request",
    format = "uuid"
)]
struct RequestId;

impl FromRequest for RequestId {
    todo!()
}

#[api_v2_operation]
async fn echo_pet_with_request_id(_: RequestId, body: web::Json<Pet>) -> web::Json<Pet> {
    unimplemented!()
}
```


macros/src/actix.rs
```diff
@@ -1069,6 +1061,147 @@ pub fn emit_v2_security(input: TokenStream) -> TokenStream {
...
+
+    let mut parameter_attrs = HashMap::new();
+
+    let valid_attrs = vec!["description", "name", "format"];
+    let invalid_attr_msg = format!(
+        "Invalid macro attribute. Should be named attribute {:?}",
+        valid_attrs
+    );
+
+    // Read security params from openapi attr.
```
tiagolobocastro, line 1095: should this be header params?

macros/src/actix.rs
```diff
@@ -1069,6 +1061,147 @@ pub fn emit_v2_security(input: TokenStream) -> TokenStream {
...
+            "binary" => quote! { Some(paperclip::v2::models::DataTypeFormat::Binary) },
+            "data" => quote! { Some(paperclip::v2::models::DataTypeFormat::Date) },
+            "datetime" => quote! { Some(paperclip::v2::models::DataTypeFormat::DateTime) },
+            "password" => quote! { Some(paperclip::v2::models::DataTypeFormat::Password) },
+            "url" => quote! { Some(paperclip::v2::models::DataTypeFormat::Url) },
+            "uuid" => quote! { Some(paperclip::v2::models::DataTypeFormat::Uuid) },
+            "ip" => quote! { Some(paperclip::v2::models::DataTypeFormat::Ip) },
+            "ipv4" => quote! { Some(paperclip::v2::models::DataTypeFormat::IpV4) },
+            "ipv6" => quote! { Some(paperclip::v2::models::DataTypeFormat::IpV6) },
+            "other" => quote! { Some(paperclip::v2::models::DataTypeFormat::Other) },
```
tiagolobocastro, line 1159: Could we add a `TryFrom<&str>` to `DataTypeFormat` ?

tiagolobocastro:
@rlebran a few questions if I may:
So we'd have one of those Apiv2Header structs for each header?

Have you considered "grouping" it, much like we currently do for the Json types?
We could also infer from the types and doc comments, example:

```rust
#[derive(Apiv2Header, Deserialize)]
struct FooHeaders {
  /// Allow to track request
  x_request_id: uuid::Uuid,
  /// Another one
  #[openapi(name = "Custom Name")]
  custom_name: i32,
}
```

The way you've done it it's an unit struct, so we can't actually get the value from it? So it's just a way to document it?

macros/src/actix.rs
rlebran: Yes it should be !

macros/src/actix.rs
rlebran: I might be wrong but if we use a `FromStr` implementation would it still fail at compile time ?

Another solution would be to use [FromMeta](https://docs.rs/darling/latest/darling/trait.FromMeta.html) but it means adding darling as a dependency.

rlebran:
> So we'd have one of those Apiv2Header structs for each header?

Currently yes, didn't think about grouping them to be honest.

> so we can't actually get the value from it? So it's just a way to document it?

We can use it like `Apiv2Security`. For example: 
```rust
    #[derive(Apiv2Header, Deserialize)]
    #[openapi(
        name = "X-Request-ID",
        description = "Allow to track request",
        format = "uuid"
    )]
    struct RequestId {
       id: uuid,
       timestamp: Date
    };```
is totally feasible, it totally depends on what you want to put in your `FromRequest` implementation.

Might not be a good idea but I saw in a code base `Apiv2Security` being use to carry a Realm information (which generate an invalid openapiv2 schema). The `FromRequest` implementation was building a state based on the Realm information.
This is still feasible if headers are grouped but it means all extractors logic have to be implemented in the same place (which is ok for me but I would prefer having your insight on that point).

File: macros/src/actix.rs (hunk 2 above)
Comment (rlebran) [reply macros/src/actix.rs:1159]: We would be better !

**Comment by tiagolobocastro** [comment]
> We can use it like `Apiv2Security`. For example:
> 
> ```rust
>     #[derive(Apiv2Header, Deserialize)]
>     #[openapi(
>         name = "X-Request-ID",
>         description = "Allow to track request",
>         format = "uuid"
>     )]
>     struct RequestId {
>        id: uuid,
>        timestamp: Date
>     };
> ```
> 
> is totally feasible, it totally depends on what you want to put in your `FromRequest` implementation.
> Might not be a good idea but I saw in a code base `Apiv2Security` being use to carry a Realm information (which generate an invalid openapiv2 schema). The `FromRequest` implementation was building a state based on the Realm information. 

Right yes `FromRequest` can build it into anything. I was just wondering if it were better to make it explicit, so the openapi is generated based on type information rather than being all "faked via proc macros", if that makes sense.

> This is still feasible if headers are grouped but it means all extractors logic have to be implemented in the same place (which is ok for me but I would prefer having your insight on that point).
If someone wants multiple ones they could always add more "groups"?
```rust
/// Headers for Foo
#[derive(Apiv2Header, Deserialize)]
struct FooHeaders {
  /// Id of foo
  foo_id: uuid::Uuid,
  /// Count of foo
  foo_count: i32,
  #[openapi(skip)]
  other_cruft: XXXX
}

/// Headers for Doo
#[derive(Apiv2Header, Deserialize)]
struct DooHeaders {
  /// id of doo
  doo_id: uuid::Uuid
}

#[api_v2_operation]
async fn handler(_: FooHeaders, _: DooHeaders, body: web::Json<X>) -> web::Json<X> {
...
}
```

If we break it down what I'm suggesting is not vastly different than what you are, it's just grouping and "auto infer" of the header name and openapi type?
We could break it down into iterations, we start for example with a simple one where the types have to be manually specified rather than inferred, example:
```rust
/// Headers for Foo
#[derive(Apiv2Header, Deserialize)]
struct FooHeaders {
  /// Id of foo
  #[openapi(
        name = "foo-id",
        description = "Id of foo",
        format = "uuid"
    )]
  foo_id: uuid::Uuid,
}
```
And then we remove add that automatically, so you only have to specify #[openapi] to override things.

That'd be my suggestion, but do let me know if your thoughts and if you think it'd even work this way or not!

rlebran:
Hey ! Sounds good to me.
I just push the first step, still need to infer `type` and `format` parameter.

rlebran:
Hey ! I think we should be close to have something working here. 
Fell free if you have any suggestion / questions / remarks.

Didn't find any good solution to handle the `DataTypeFormat` matching in the macro... Running out of ideas on this one

tiagolobocastro:
> Hey ! I think we should be close to have something working here. Fell free if you have any suggestion / questions / remarks.
> 
> Didn't find any good solution to handle the `DataTypeFormat` matching in the macro... Running out of ideas on this one

Sorry I've been a bit busy...
But it's looking great! As for the `DataTypeFormat` it's fine as it is then, maybe later we can figure out some other day.

rlebran:
No problem ! From what I saw, only `api_v2_schema_struct` is used in the core. The other use of the macro crate is to re-expose the different macros.
One way would be to split the macro crate into 2 so core would be able to become a dependency of macros and it would be easier to do the mappings (and other stuff like relying on the `SPECIAL_HEADERS` const for example). 
But if it's something that needs to be done, I think it deserve it's own PR

tiagolobocastro:
> No problem ! From what I saw, only `api_v2_schema_struct` is used in the core. The other use of the macro crate is to re-expose the different macros. One way would be to split the macro crate into 2 so core would be able to become a dependency of macros and it would be easier to do the mappings (and other stuff like relying on the `SPECIAL_HEADERS` const for example). But if it's something that needs to be done, I think it deserve it's own PR

If that's possible that would be great, and yes definitely on another PR :)

macros/src/actix.rs
```diff
@@ -1069,6 +1061,210 @@ pub fn emit_v2_security(input: TokenStream) -> TokenStream {
...
+    fn quote_format(format: String) -> proc_macro2::TokenStream {
+        match &*format {
+            "int32" => quote! { Some(paperclip::v2::models::DataTypeFormat::Int32) },
+            "int64" => quote! { Some(paperclip::v2::models::DataTypeFormat::Int64) },
+            "float" => quote! { Some(paperclip::v2::models::DataTypeFormat::Float) },
+            "double" => quote! { Some(paperclip::v2::models::DataTypeFormat::Double) },
+            "byte" => quote! { Some(paperclip::v2::models::DataTypeFormat::Byte) },
+            "binary" => quote! { Some(paperclip::v2::models::DataTypeFormat::Binary) },
+            "data" => quote! { Some(paperclip::v2::models::DataTypeFormat::Date) },
+            "datetime" => quote! { Some(paperclip::v2::models::DataTypeFormat::DateTime) },
```
tiagolobocastro, line 1104: this should be date-time (with the dash in the middle) ?

macros/src/actix.rs
```diff
@@ -1069,6 +1061,210 @@ pub fn emit_v2_security(input: TokenStream) -> TokenStream {
...
+                    use paperclip::v2::schema::TypedData;
+                    #ty_ref::format()
+                } },
+            )
+        } else {
+            (quote! { None }, quote! { None })
+        };
+
+        let (quoted_type, quoted_format) = if let Some(format) = parameter_attrs.get("format") {
+            let quoted_format = quote_format(format.clone());
```
tiagolobocastro, line 1223: `quote_format` could take `&str` and then we don't need to `.clone()` here?

macros/src/actix.rs
```diff
@@ -1069,6 +1061,210 @@ pub fn emit_v2_security(input: TokenStream) -> TokenStream {
...
+
+        header_definitions.push(def_block);
+    }
+
+    let gen = quote! {
+        impl #impl_generics paperclip::v2::schema::Apiv2Schema for #name #ty_generics #where_clause {
+            fn header_parameter_schema() -> Vec<paperclip::v2::models::Parameter<paperclip::v2::models::DefaultSchemaRaw>> {
+                let mut headers = vec![];
+                #(headers.push(#header_definitions));*;
+                headers
```
tiagolobocastro, line 1250: Could we have something like this:
```suggestion
                vec![
                    #(#header_definitions),*
                ]
```

tiagolobocastro (commented):
Great stuff!
Could you please add something to the changelog as well?

macros/src/actix.rs
rlebran: I did added both (`"datetime" | "date-time"`) because serde rename it to `date-time` but the `ToString` implementation return `datetime`

rlebran:
There is one last thing I just saw and fixed.
What do you want to do with this case: 
```rust
#[derive(Apiv2Header, Deserialize)]
#[openapi(
    name = "X-Request-ID",
    description = "Allow to track request",
    format = "uuid"
)]
struct RequestId;```
Currently `openapi` the macro emit an error indicating that the `openapi` attribute is miss placed and should be defined at field level. Sounds good ?

**Comment by tiagolobocastro** [comment]
> There is one last thing I just saw and fixed. What do you want to do with this case:
> 
> ```rust
> #[derive(Apiv2Header, Deserialize)]
> #[openapi(
>     name = "X-Request-ID",
>     description = "Allow to track request",
>     format = "uuid"
> )]
> struct RequestId;
> ```
> 
> Currently `openapi` the macro emit an error indicating that the `openapi` attribute is miss placed and should be defined at field level. Sounds good ?

Yep sounds good!

File: CHANGELOG.md [hunk 6]
```diff
@@ -4,8 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.7.1] - 2022-04-10
+## [0.8.0] - 2022-04-19
```
tiagolobocastro, line 7: could you replace this line with `## [Unreleased]` (just to follow what's been done till now).
Btw, there's no breaking change right, we'll be able to release a minor?

tiagolobocastro (approved):

CHANGELOG.md
rlebran: No there is no breaking change but looking at the changelog it seems that every addition increase the minor

tiagolobocastro:
Hey @rlebran sorry I've been very busy and haven't been able to take a look at these. Is this PR ready to merge?

rlebran:
Hey ! Yes, this PR is ready to merge

Issue #404, opened 2022-03-22
Lost compatibility with `actix_web_httpauth`


ManevilleF:
Using `actix_web` 4 and paperclip as a git dependency with the new `actix4` feature:

```toml
[dependencies.paperclip]
version = "0.6"
git = "https://github.com/paperclip-rs/paperclip"
branch = "master"
features = ["actix4", "v3", "swagger-ui", "actix-files"]
```

The `paperclip_actix::web::Resource::wrap` method no longer supports the  `actix_web_httpauth::middleware::HttpAuthentication::bearer` middleware.

I get the following error when using it like:

```rust
.service(web::resource("/route/{id}")
         .wrap(HttpAuthentication::bearer(bearer_token_middleware))
         .route(web::get().to(my_controller::show))
)
```

```
error[E0271]: type mismatch resolving `<HttpAuthentication<BearerAuth, fn(ServiceRequest, BearerAuth) -> impl std::future::Future<Output = Result<ServiceRequest, actix_web::Error>> {bearer_token_middleware}> as Transform<actix_web::resource::ResourceService, ServiceRequest>>::Response == ServiceResponse`
   --> file.rs
    |
94  |                     .wrap(HttpAuthentication::bearer(bearer_token_middleware))
    |                      ^^^^ expected struct `BoxBody`, found enum `EitherBody`
    |
    = note: expected struct `ServiceResponse<BoxBody>`
               found struct `ServiceResponse<EitherBody<_>>`
note: required by a bound in `paperclip::actix::web::Resource::<actix_web::Resource<T>>::wrap`
   --> /me/.cargo/git/checkouts/paperclip-30ddbbc2fa493bf0/a0e9d32/plugins/actix-web/src/web.rs:178:17
    |
178 |                 Response = ServiceResponse,
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `paperclip::actix::web::Resource::<actix_web::Resource<T>>::wrap`
    ```

tiagolobocastro:
Tagging @sfisol in case he's able to take a look

robjtede:
Paperclip either needs to wrap everything passed to `.wrap` in `Compat` or support non-`BoxBody` response payload types properly. This is not strictly a blocker to releasing v4 support since the workaround is for users to use `Compat`.

tiagolobocastro:
Thanks @robjtede - yeah `Compat` is fine as a workaround for now.
I'm preparing a new release now.

[...]

PR #414, opened 2022-04-09
Fix issue #404

rlebran:
This should fix issue #404 by properly supporting non-BoxBody response payload types.


tiagolobocastro:
Awesome, thanks @rlebran ! 
Would you mind adding this to the changelog.md?
Also do you mind adding the serde flatten fix and the "ref URIs should be RFC3986-compliant" fix as well? 
Otherwise I can add it on a separate PR, just let me know, thanks!

rlebran:
Hey! Adding it to the changelog as soon as I have my computer close by. Should be in few hours but you can do it I a separate PR if you want to

tiagolobocastro:
thanks @rlebran 🥳

Issue #417, opened 2022-04-19
ability to set example for string responses


Tricked-dev:
There's no way to change the example from a string i suggest that there should be a way to do this

```rs
#[derive(Apiv2Schema, Serialize, Deserialize, Clone)]
pub struct Example {
  /// Foo
  /// * bar
  pub foo: String,
  // or
  /// Foo
  #[openapi(example = "bar"]
  pub foo: String
}
```

The same would apply to to bool/int keys. Ints currently default to 0 and booleans to true.

rlebran:
Hey! I was looking at it few days ago but didn't commit anything yet. Might have something pretty soon thought

dunnock:
This issue seems complete?

rlebran:
Yes, this issue should be fixed

PR #421, opened 2022-04-23
Feat: add example to schema definition

rlebran:
Add example for schema and fields through openapi attribute. 

This produce valid openapi specification. I changed the example type from `String` to `serde_json::Value` because the specification describe example as 

> A free-form property to include an example of an instance for this schema.

and examples show a json like example value.

This is link to #417


tiagolobocastro (approved):
Thanks @rlebran!
Btw, what do you think about using the Default trait as another way of specifying examples automatically?
Maybe something like `#[openapi(default_examples)]`.

rlebran:
That's a good idea but should it be opt-in or if a Default implementation exists we can use it automatically ?

PR #423, opened 2022-05-07
Feat undocumented routes

rlebran:
Hi,

Using paperclip, we currently struggle to generate openapi schema without some routes. 
Due to Actix scope handling (can't currently define twice the same scope with different route) we have no way to have undocumented routes sharing a path prefix with documented ones.

This PR add a `skip` attribute to `api_v2_schema` macro allowing the user to have some undocumented route. Made this opt-in so user don't accidentally forget to document some route.


rlebran:
I will take a look at that clippy failure when I have some time

tiagolobocastro (approved):

PR #428, opened 2022-05-15
Handle parameters on the format r#type

snaggen:
When you have a struct with  a member using a reserved word as a name, it will be specified as r#type. For Query structs, this is probably quite common. Previously they got hte name "r#type" in the openapi spec, so I just added a strip_prefix r# to handle this case.


snaggen:
This did not work at all... flawed tests gave me the wrong imression

snaggen:
Updated with working fix

PR #429, opened 2022-05-15
Handle parameters on the format r#type

snaggen:
A new version of my previous non working PR, this seems to actually do what I intended.

When you have a struct with a member using a reserved word as a name, it will be specified as r#type. For Query structs, this is probably quite common. Previously they got hte name "r#type" in the openapi spec, so I just added a strip_prefix r# to handle this case.

### DIFF
--- a/macros/src/actix.rs
+++ b/macros/src/actix.rs
@@ -1542,6 +1542,9 @@ fn handle_field_struct(
             .expect("missing field name?")
             .to_string();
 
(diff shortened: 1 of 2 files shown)


rlebran:
Hi, is it possible to add a test so we prevent any regressions?

snaggen:
Test added

tiagolobocastro (approved):

Issue #427, opened 2022-05-13
paths with regex like /{path:.*} gets the regex included in the name


snaggen:
When you specify a path that includes a regex, like "/{path:.*}", then the regex part will be included in the name in the spec, like ```"name":"path:.*"```, I would have expected the name in this case to just be the part before the ':', which would be "path" in this case.

PR #430, opened 2022-05-15
Clean name from template.

snaggen:
Avoid including regex from the path template in the name
Fixes: #427


snaggen:
Added test

tiagolobocastro (approved):

PR #435, opened 2022-05-31
upgrade semver

matklad:
semver 1.0 is a much leaner, better crate than it was prior to 1.0. Now
it is stable and will unlikely to see further changes.

We use semver privatly, so this bump is not a semver change for us.


tiagolobocastro (approved):
thanks @matklad

PR #447, opened 2022-07-06
fix(deps): Update Uuid dep

LeoTayot:
Update Uuid dep to prevent the following error when using a more recent version of uuid crate.

note :
I got this error simply while using latest version of uuid dep in an schema :
```rust
#[derive(Apiv2Schema)]
pub struct Pet {
    pub id: Uuid, // Triggered the error
    pub name: String,
}
```
---
image


tiagolobocastro:
Handled by #461

PR #455, opened 2022-08-03
Select a release series for uuid which is compatible with diesel

saethlin:
`uuid = "0"` permits implicit/automatic breaking updates. This version requirement has been the source of many bizarre compilation issues for us, because `paperclip-core` will select `uuid` version 0.8.2, but diesel only provides impls on `uuid` version 0.6.5.

A carefully-manicured `Cargo.lock` can get us out of this situation, but it ends up being a pretty janky developer experience, because any accidental changes to it can often break the build with confusing errors.

The big downside to this is that it puts `paperclip` on `uuid` version 0.6, which in my experience is the only way to make things work with `diesel`. I see that you've previously been using 0.8 (in development?). Is there any particular reason why?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ actix-web3 = { version = "3", default-features = false, package = "actix-web" }
 actix-web4 = { version = "4", default-features = false, package = "actix-web" }
 chrono_dev = { version = "0.4", features = ["serde"], package = "chrono" }
 futures = "0.3"
-uuid_dev = { version = "0.8", features = ["serde"], package = "uuid" }
+uuid_dev = { version = "0.6", features = ["serde"], package = "uuid" }
 reqwest = { version = "0.10", features = ["blocking", "json"] }
 log = { version = "0.4", features = ["kv_unstable"] }
 insta = "1.0"
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -30,7 +30,7 @@ serde = { version = "1.0", features = ["derive", "rc"] }
 serde_json = "1.0"
 serde_yaml = "0.8"
 url = { version = "2", optional = true }
-uuid = { version = "0", optional = true }
+uuid = { version = "0.6", optional = true }
 thiserror = "1.0"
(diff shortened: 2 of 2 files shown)


tiagolobocastro:
hmm I can't give you an answer as to why it's set to 0.8, seems it's been this way for years.
We also have another PR to bump it to 1, see #447.

Honestly I'm not sure what to do with this if different users have different requirements?
Provide different uuid features? Egg, uuid0 and uuid1? 
Doesn't seem great, but not sure what else to do in order to support both, what do you think?

saethlin:
I would prefer a `uuid0` and `uuid1` feature.

The situation as it is now is extremely confusing to new Rust users, so I don't think making users pick one of the features is a big problem.

weiznich:
Diesel maintainer here. Let me clarify a few things, as there is some misconception in the OP.

> but diesel only provides impls on uuid version 0.6.5.

This is not true for recent diesel versions. Diesel 1.4.8 support uuid with the versions between 0.2.x and 0.6.x behind the `uuid` feature flag **and** uuid 0.7.x and 0.8.x behind the `uuidv07` feature flag. For the upcoming 2.0 release and the recent rc releases any uuid version > 0.7, <2.0 is supported behind the `uuid` feature flag.

> uuid = "0" permits implicit/automatic breaking updates.

Using `uuid = "0"` as version constraint allows compatibility with all uuid versions supported by diesel. Cargo will just not select the right one by default in some cases. That's something you as a user can fix by issuing `cargo update -p uuid --precise=some.version.path` where the version needs to be supported by all of your crates. To be clear that's a bug/issue in how cargo resolves dependencies. There is no real way for crates to fix that.

tiagolobocastro:
Thank you @weiznich, so in that case I wonder if we could simply have a uuid0 feature which uses uuid="0" and a uui1 with uses uuid="1" ? Paperclip itself doesn't have any constraints on the uuid, we just impl a trait for the type, with no self.

saethlin:
I think between the PR referenced above and the new diesel release, this has been taken care of in better ways. Thanks all!

PR #457, opened 2022-08-06
fix: support non-boxed bodies in scope middleware

jtscuba:
This PR makes paperclip's scope compatible with middlewares that don't return a box body. See the issue I opened here.
 https://github.com/paperclip-rs/paperclip/issues/456.

Changes
1. Add an extra type parameter `B` with the same constraints as `actix_web::Scope` to the wrap function on paperclip's `Scope`
2. Remove the `Response = ServiceResponse` constraint on the impl block of `Scope<actix_web<Scope<T>>>`
3. Updated the `Response = ServiceResponse` constraint on route to `Response = <<actix_web::Route as ServiceFactory<ServiceRequest>>::Service as actix_service::Service<ServiceRequest,>>::Response`, AKA the response type of route as specified by actix-web. This happens to be `ServiceResponse` as of 4.1.0 so not an actual change but it does remove a place paperclip is hard-coding the ServiceResponse type with it's implicit BoxBody parameter. This change isn't required so open to removing it.

Testing
1. Tested locally against my own project [discord dungeon](https://discorddungeon.com)
2. Added two middlewares to the configure block of the `test_simple_app` test case, verified the tests failed on master and pass on my branch.

### DIFF
--- a/plugins/actix-web/src/web.rs
+++ b/plugins/actix-web/src/web.rs
@@ -283,15 +283,16 @@ impl Scope {
     }
 }
 
-impl<T> HttpServiceFactory for Scope<actix_web::Scope<T>>
+impl<T, B> HttpServiceFactory for Scope<actix_web::Scope<T>>
 where
     T: ServiceFactory<
             ServiceRequest,
             Config = (),
-            Response = ServiceResponse,
+            Response = ServiceResponse<B>,
             Error = Error,
             InitError = (),
         > + 'static,
+    B: MessageBody + 'static,
 {
     fn register(self, config: &mut AppService) {
         if let Some(s) = self.inner {
@@ -302,13 +303,7 @@ where
 
 impl<T> Scope<actix_web::Scope<T>>
 where
-    T: ServiceFactory<
-        ServiceRequest,
-        Config = (),
-        Response = ServiceResponse,
-        Error = Error,
-        InitError = (),
-    >,
+    T: ServiceFactory<ServiceRequest, Config = (), Error = Error, InitError = ()>,
 {
     /// Proxy for [`actix_web::Scope::guard`](https://docs.rs/actix-web/*/actix_web/struct.Scope.html#method.guard).
     ///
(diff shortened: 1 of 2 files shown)


jtscuba:
@tiagolobocastro

tiagolobocastro:
@jtscuba on the issue you mentioned actix3 but it seems now you're making the changes for actix4 only, is this right?
Could you please fix the rustfmt changes? thanks

jtscuba:
@tiagolobocastro That's correct. Originally I thought web3 went with openapi v3 but it actually goes with actix-web 3. The definitions are fine as-is for actix-web 3, actix-web added the new type parameter in v4

tests/test_app.rs
```diff
@@ -182,16 +183,23 @@ fn test_simple_app() {
...
 
     fn config(cfg: &mut web::ServiceConfig) {
+        let test_scope = web::scope("/test")
+            .service(web::resource("{id}").route(web::get().to(path_with_param_without_slash)));
+
+        // Non-boxed middlewares are only supported in actix 4, prevously use of the Compat middleware was required
+        #[cfg(feature = "actix4")]
+        let test_scope = test_scope
+            .wrap(Logger::default())
+            .wrap(DefaultHeaders::default().add(("X-Test", "Value")));
```
tiagolobocastro, line 193: would you mind creating a new test just for actix4, can be very simple, just testing this works, example: `test_non_boxed_bodies` ?

plugins/actix-web/src/web.rs
```diff
@@ -521,6 +517,10 @@ impl ServiceFactory<ServiceRequest> for Route {
     type InitError = ();
     type Service = <actix_web::Route as ServiceFactory<ServiceRequest>>::Service;
     type Future = <actix_web::Route as ServiceFactory<ServiceRequest>>::Future;
+    // type Response =
```
tiagolobocastro, line 520: Do we need this as well?

jtscuba: Ah yes, this is the removal of the hardcoded reference to ServiceResponse I mentioned. Forgot to comment that back in.

jtscuba:
@tiagolobocastro updated.

tiagolobocastro (approved):

PR #458, opened 2022-08-24
fix: Fixed header-based SecuritySchema conversion for OpenAPI v3

frol:


tiagolobocastro:
@frol I think this is a good fix, we'll have to expose some other way of selecting the http scheme which is not present in 2.0.
Could you please fix the lint failure?

frol:
@tiagolobocastro Done

tiagolobocastro (approved):

PR #461, opened 2022-08-27
feat: add uuid0 and uuid1 features

tiagolobocastro:
Add a way to specify uuid 0 or 1, or both!

Signed-off-by: Tiago Castro <tiagolobocastro@gmail.com>


tiagolobocastro:
Based on #455 and #447 we exposed both version 0 and 1.
There is no nice solution here... PTAL

PR #463, opened 2022-08-31
feat: Respect `host` setting of v2 spec when converting to v3

frol:
`schemes` and `host` settings used to be completely ignored. This PR properly handles those when populating v3 `servers` field.

### DIFF
--- a/core/src/v3/openapi.rs
+++ b/core/src/v3/openapi.rs
@@ -6,7 +6,7 @@ impl From<v2::DefaultApiRaw> for openapiv3::OpenAPI {
             openapi: "3.0.0".into(),
             tags: v2.tags.iter().cloned().map(From::from).collect(),
             info: v2.info.clone().into(),
-            servers: openapi3_server(v2.host, v2.base_path),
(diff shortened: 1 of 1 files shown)


core/src/v3/openapi.rs
```diff
@@ -55,15 +55,46 @@ impl From<v2::DefaultApiRaw> for openapiv3::OpenAPI {
...
+                        url: format!("{}://{}{}", scheme_str, host, base.as_deref().unwrap_or("")),
+                        description: None,
+                        variables: None,
+                        extensions: indexmap::IndexMap::new(),
+                    }
+                })
+                .collect()
+        } else {
+            vec![openapiv3::Server {
+                url: format!("//{}{}", host, base.as_deref().unwrap_or("")),
```
tiagolobocastro, line 86: Should this be:
`url: format!("{}{}", host, base.as_deref().unwrap_or("")),` ?

frol: Nope, if you do that way it will be a relative path while this URL (starting with //) uses the “current” protocol as per the v2 specification when the scheme is omitted

PR #467, opened 2022-09-30
Add Content-Type Header to Swagger-UI Requests

mikemcdermottredjack:


plugins/actix-web/Cargo.toml
```diff
@@ -18,6 +18,7 @@ actix-service2 = { version = "2", package = "actix-service", optional = true }
 actix-web2 = { version = "2", default-features = false, optional = true, package = "actix-web" }
 actix-web3 = { version = "3", default-features = false, optional = true, package = "actix-web" }
 actix-web4 = { version = "4", default-features = false, optional = true, package = "actix-web" }
+mime_guess = { version = "2.0.4" }
```
tiagolobocastro, line 21: Do we need "rev-mappings"? If not could we use `default-features = false` ?

tiagolobocastro:
Sorry @mikemcdermottredjack just fixed the failing tests on master, would mind rebasing?

PR #472, opened 2022-10-16
Replaced dependency `pin-project` with `pin-project-lite`

xilec:
`actix-web` and `tokio` have no dependency on `pin-project` (using more lightweight version `pin-project-lite`). I suppose it would be good to support this appoach.


tiagolobocastro (approved):
Thanks, looks good to me!

tiagolobocastro:
@xilec could you please take a look at the test failures? Thanks

PR #484, opened 2022-12-08
Set refresh_url to an empty string

wzzrd:
The previous value of refresh_url (None) translated into 'null' in the spec, which is not accepted as a proper value, as refresh_url is supposed to have a string value.

To see an example of how this goes wrong, you can create a spec with the old code (i.e. with refresh_url: None) and paste it into editor.swagger.io. You'll get an error. Now generate a spec using this
new code (i.e. with refresh_url: "") and the error goes away.

### DIFF
--- a/core/src/v3/security_scheme.rs
+++ b/core/src/v3/security_scheme.rs
@@ -31,14 +31,14 @@ impl From<v2::SecurityScheme> for openapiv3::SecurityScheme {
                         implicit: match flow.as_str() {
                             "implicit" => Some(openapiv3::OAuth2Flow::Implicit {
                                 authorization_url: v2.auth_url.clone().unwrap_or_default(),
-                                refresh_url: None,
+                                refresh_url: Some("".to_string()),
                                 scopes: scopes.clone(),
                             }),
                             _ => None,
                         },
                         password: match flow.as_str() {
                             "password" => Some(openapiv3::OAuth2Flow::Password {
-                                refresh_url: None,
+                                refresh_url: Some("".to_string()),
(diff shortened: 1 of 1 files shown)


tiagolobocastro:
@wzzrd seems like this should be raised against the openapiv3 crate if an `Option` is not a valid type to use here?

wzzrd:
Fair point. I just looked into why None doesn't result into the refresh_url being omitted for serialization and it looks like the skip_serializing_if option was only added to this field a couple of months ago, in a yet to be tagged version of their crate.

I've asked the team of the openapiv3 crate to tag a release that includes that fix. Hopefully you can release a new paperclip version that pulls in that new update? Thanks for this awesome crate, btw! :)

wzzrd:
I'll close this PR and open a new one when openapiv3 releases a fixed version of their crate

tiagolobocastro:
> Fair point. I just looked into why None doesn't result into the refresh_url being omitted for serialization and it looks like the skip_serializing_if option was only added to this field a couple of months ago, in a yet to be tagged version of their crate.
> 
> I've asked the team of the openapiv3 crate to tag a release that includes that fix. Hopefully you can release a new paperclip version that pulls in that new update? Thanks for this awesome crate, btw! :)

Absolutely, once the PR gets merged I'll create a release.
Thanks but props goes for @wafflespeanut! :)

wzzrd:
Brilliant, the openapiv3 team already released v1.0.2: https://github.com/glademiller/openapiv3/commit/bdd22f046d2bc19ede257504645d31f835545222

That should hopefully fix the refresh_url problem.

@tiagolobocastro thanks!

Issue #501, opened 2023-04-04
[Bug] Type mismatch


AzHicham:
Hello, 

I'm currently facing a sever bug while using Paperclip. 
Having this code : 

```
pub fn config<T: ServeSlide + 'static>() -> Scope {
    web::scope("/imgsrv")
        .route("/roi/{path:.*}+{level}", web::get().to(roi::<T>))
        ...
}
```
```
#[api_v2_operation]
pub async fn roi<T: ServeSlide>(
    input: web::Path<RoiRequestInput>,
) -> Result<Json<Vec<Roi>>, T::ErrorType> {
   ...
    Ok(Json(res))
}
```
```
#[derive(Apiv2Schema, Debug, Deserialize)]
pub struct RoiRequestInput {
    pub path: String,
    pub level: u32,
}
```
I get the following generated openapi schema: 

```
"/imgsrv/roi/{path}+{level}": {
      "get": {
        "responses": {
          "200": {
            "description": "OK",
            "schema": {
              "type": "array",
              "items": {
                "$ref": "#/definitions/Roi"
              }
            }
          }
        },
        "parameters": [
          {
            "in": "path",
            "name": "path",
            "required": true,
            "type": "integer",
            "format": "int32",
            "pattern": ".*"
          },
          {
            "in": "path",
            "name": "level",
            "required": true,
            "type": "string"
          }
        ]
      }
    },
```

As you can see paperclip switch integer type with string 

Thanks for your help

jonra1993:
I have a similar error in my path variables not have correct data

```rust
#[derive(Deserialize, Apiv2Schema)]
struct PathInfo {
    user_id: uuid::Uuid,
    friend: String,    
}

#[derive(Serialize, Apiv2Schema)]
struct IPostResponse<T> {
    message: String,
    data: T,
}

#[api_v2_operation]
#[get("/users/{user_id}/{friend}")] // <- define path parameters
async fn get_user(
    info: Path<PathInfo>,
    query: Query<QueryInfo>,
) -> Result<Json<IPostResponse<String>>, Error> {
    let result = if let Some(username) = &query.username {
        format!(
            "Welcome {}, user_id {}, username {}! ",
            info.friend, info.user_id, username
        )
    } else {
        format!("Welcome {}, user_id {}! ", info.friend, info.user_id)
    };
    let response = IPostResponse {
        data: result,
        message: String::from("Data got sucessfully"),
    };
    Ok(web::Json(response))
}
```

[...]

PR #504, opened 2023-08-02
fix(path-order): don't trash previous parameter names

tiagolobocastro:
There's a routine which patches parameter names and it trashes previous attempts when it runs twice.
I'm not sure this is meant to work but it seems that the only time we need this function is when we don't have param names, example:
_p: web::Path<(bool, u32, String)>

We seem to completely ignore param names and just name things from the path, example:
get(ace: web::Path<bool>, user_id: web::Path<u32>, friend: web::Path<String>) This can actually be written as
get(user_id: web::Path<bool>, ace: web::Path<u32>, friend: web::Path<String>)

Which actually swaps parameters in the generate json!!

Resolves: #501

### DIFF
--- a/core/src/v2/actix.rs
+++ b/core/src/v2/actix.rs
@@ -568,7 +568,7 @@ macro_rules! impl_path_tuple ({ $($ty:ident),+ } => {
                         description: v.description,
                         collection_format: None, // this defaults to csv
                         items: v.items.as_deref().map(map_schema_to_items),
-                        name: k,
+                        name: String::new(),
                         ..Default::default()
                     }));
                 }
--- a/tests/test_app.rs
+++ b/tests/test_app.rs
@@ -141,6 +141,220 @@ impl Responder for Pet {
 }
 
 #[test]
+fn test_path_order() {
+    #[derive(Deserialize, Apiv2Schema)]
+    #[allow(unused)]
+    struct PathInfo {
+        friend: String,
+        user_id: u32,
+        ace: bool,
+    }
+    #[api_v2_operation]
+    #[get("/struct/users/{ace}/{user_id}/{friend}")]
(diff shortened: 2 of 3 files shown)


core/src/v2/actix.rs
```diff
@@ -568,7 +568,7 @@ macro_rules! impl_path_tuple ({ $($ty:ident),+ } => {
                         description: v.description,
                         collection_format: None, // this defaults to csv
                         items: v.items.as_deref().map(map_schema_to_items),
-                        name: k,
+                        name: String::new(),
```
mharkins-cosm, line 571: I don't know the code very well at all. Does it matter at all that `k` is used for determining whether the parameter is "required", but then discarding it when determining the parameter "name"?  Is it possible we end up marking the wrong parameters as "required"?

tests/test_app.rs
```diff
@@ -140,6 +140,220 @@ impl Responder for Pet {
     }
 }
 
+#[test]
+fn test_path_order() {
```
mharkins-cosm, line 144: I love this test! Does it fail without the proposed changes?

core/src/v2/actix.rs
tiagolobocastro: Not here, this is the impl for path tuples: Path<($($ty,)+)
So here the names are just indexes, "0", "1"....

But even with individual paths, so two Path<>, Path<>, the names are not matched with the url names, example:
/{continent}/{country}
fn f(country: Path<String>, continent: Path<String>)...

Here as you see, country and continent are swapped, but no error is thrown, in fact the arg names in the function seem to be completely ignored..
Fixing this will require a lot more changes probably.

tests/test_app.rs
tiagolobocastro: Yes it does.
Though after writing it I found out we also have another one: test_params, which was not failing when it should have! but it's also a more complicated test.

PR #514, opened 2023-10-12
add lint ignores to paperclip unit struct

formlogic-kirk:
Code generated from the `api_v2_operation` proc macro includes a unit struct definition of the form `paperclip_snake_case_struct_name` and was trigging the non-camel-case type name lint. Added the same lint ignores here as are used on the other struct definition in the same macro.


tiagolobocastro (approved):

tiagolobocastro:
thanks @formlogic-kirk

PR #525, opened 2024-06-30
build: update openapiv3 to v2.0

tiagolobocastro:
There's a breaking change on the crate which does not expose the fields as pub and so we need to use our own patch for now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ url_dep = { version = ">=1.7,<3", package = "url" }
 thiserror = "1.0"
 anyhow = "1.0"
 once_cell = "1.4"
-openapiv3 = { version = "1.0.3", optional = true }
+openapiv3-paper = { version = "2.0", optional = true }
 
 [dev-dependencies]
 actix-rt1 = { version = "1.0", package = "actix-rt" }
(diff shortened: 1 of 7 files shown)


tiagolobocastro:
Raised PR on openapiv3: https://github.com/glademiller/openapiv3/pull/86

tiagolobocastro:
The raised PR is still not reviewed, I wonder if the repo is no longer maintained.
So we'd have to consider forking or switching to another crate in order to make progress here and in the other PR which is depending on this one...
@Sufflope @formlogic-kirk @snaggen @RaulTrombin @thomedw what are your thoughts here?

thomedw:
Hi @tiagolobocastro, 

There's a crate called [openapiv3-extended](https://crates.io/crates/openapiv3-extended), which is a fork of the repository you mentioned. It might be worth checking out.

tiagolobocastro:
> Hi @tiagolobocastro,
> 
> There's a crate called [openapiv3-extended](https://crates.io/crates/openapiv3-extended), which is a fork of the repository you mentioned. It might be worth checking out.
> 
> I think it has the required pub fields https://github.com/kurtbuilds/openapiv3/blob/master/src/security_scheme.rs#L67.

I did not know this, thank you! I'll take a look

tiagolobocastro:
Seems like updating to the extended crate would take a few changes, and we might have to raise a PR there as well.
To avoid going through this loop again, I've added a new crate, a simple fork with just these changes.
Since openapiv3 is not part of the public api yet, this should be fairly harmless but we do need to consider how to progress forward.

PR #527, opened 2024-09-03
switch to const generic array instead of 0-32 manual impl of schema type

formlogic-kirk:
just reduces a bit of code and macro codegen

from `cargo-msrv`
master: `Finished The MSRV is: 1.65.0  `
After: `Finished The MSRV is: 1.65.0 `


tiagolobocastro (approved):
Great change, thank you!

PR #531, opened 2024-09-07
fix(actix): support parsing arrays for schemas

### LINKED ISSUE #454: Arrays are not supported
With version 0.7.1:

i have a response struct like:

```rust
#[derive(Debug, Serialize, Clone, Apiv2Schema)]
struct TestResponse {
      // Some data
      position: [f32; 3],
}
```
With a fixed size array field called `position`. When I build my application I get the following errors:

```
error[E0425]: cannot find crate `raw_schema` in the list of imported crates
 --> my_crate/src/responses/test.rs
  |
6 | #[derive(Debug, Serialize, Clone, Apiv2Schema)]
  |                                   ^^^^^^^^^^^ not found in the list of imported crates
  |
  = note: this error originates in the derive macro `Apiv2Schema` (in Nightly builds, run with -Z macro-backtrace for more info)
error[E0425]: cannot find crate `required` in the list of imported crates
 --> my_crate/src/responses/test.rs
  |
6 | #[derive(Debug, Serialize, Clone, Apiv2Schema)]
  |                                   ^^^^^^^^^^^ not found in the list of imported crates
  |
  = note: this error originates in the derive macro `Apiv2Schema` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Removing the `position` field fixes compilation

tiagolobocastro:
Array field types were not being parsed.
Also in this case errors were not being handled gracefully.

Resolves #454

### DIFF
--- a/macros/src/actix.rs
+++ b/macros/src/actix.rs
@@ -1443,7 +1443,9 @@ fn add_optional_impl(name: &Ident, generics: &Generics) -> proc_macro2::TokenStr
 
 fn get_field_type(field: &Field) -> Option<proc_macro2::TokenStream> {
     match field.ty {
-        Type::Path(_) | Type::Reference(_) => Some(address_type_for_fn_call(&field.ty)),
+        Type::Path(_) | Type::Reference(_) | Type::Array(_) => {
+            Some(address_type_for_fn_call(&field.ty))
+        }
         _ => {
             emit_warning!(
                 field.ty.span().unwrap(),
@@ -1491,7 +1493,10 @@ fn handle_unnamed_field_struct(
                 continue;
             }
 
-            let ty_ref = get_field_type(field);
+            let ty_ref = match get_field_type(field) {
+                Some(ty_ref) => ty_ref,
+                None => continue,
+            };
 
             let docs = extract_documentation(&field.attrs);
             let docs = docs.trim();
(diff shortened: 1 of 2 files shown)


tiagolobocastro:
@sireaev this should fix it.
@formlogic-kirk seems arrays were not being handled properly, would you mind taking a look at this fix?

macros/src/actix.rs
```diff
@@ -1491,7 +1493,10 @@ fn handle_unnamed_field_struct(
                 continue;
             }
 
-            let ty_ref = get_field_type(field);
+            let ty_ref = match get_field_type(field) {
```
formlogic-kirk, line 1496: good place for `let-else` if that's allowed by the msrv

formlogic-kirk (approved):
That looks good to me. I need to revisit the maps and see if this is happening for the value types there also

macros/src/actix.rs
tiagolobocastro: hmm seems that it depends on the imported crates
With my lockfile it's 1.60 but if I delete it it's 1.65 due to:

┌─────────────────────────────────────────────────────────────────────
│ error: package `hashbrown v0.14.5` cannot be built because it requires rustc 1.63.0 or newer, while the currently active rustc version is 1.62.1                                                                                         │
│ error: package `serde_yaml v0.9.34+deprecated` cannot be built because it requires rustc 1.64 or newer, while the currently active rustc version is 1.63.0                                                                               │
│ error: package `regex-automata v0.4.7` cannot be built because it requires rustc 1.65 or newer, while the currently active rustc version is 1.64.0                                                                                       │
└─────────────────────────────────────────────────────────────────────
```

formlogic-kirk: Fair enough. The MSRV aware resolver can't come soon enough

PR #535, opened 2024-09-10
doc: Add instructions for git submodules

RaulTrombin:
Solves:
https://github.com/paperclip-rs/paperclip/issues/534

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ prepare:
 	rustup component add rustfmt
 	rustup component add clippy
 	rustup toolchain install nightly --allow-downgrade -c rustfmt clippy
+	git submodule update --init --recursive
 
 check:
 	cargo +nightly fmt --all
--- a/README.md
+++ b/README.md
@@ -16,14 +16,36 @@ It's currently under active development and may not be ready for production use 
 
 You may be interested in:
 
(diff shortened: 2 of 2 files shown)


tiagolobocastro:
What about the previous suggestion of adding the git command to the make prepare command? Do you think it's better to keep as is?

RaulTrombin:
> What about the previous suggestion of adding the git command to the make prepare command? Do you think it's better to keep as is?

This would also resolve the issue. I tested it on a clean clone of the repo, and it seems fine. May I go ahead and add this commit as well?

tiagolobocastro:
> > What about the previous suggestion of adding the git command to the make prepare command? Do you think it's better to keep as is?
> 
> This would also resolve the issue. I tested it on a clean clone of the repo, and it seems fine. May I go ahead and add this commit as well?

Yeah I think so if you don't mind

RaulTrombin:
> > > What about the previous suggestion of adding the git command to the make prepare command? Do you think it's better to keep as is?
> > 
> > 
> > This would also resolve the issue. I tested it on a clean clone of the repo, and it seems fine. May I go ahead and add this commit as well?
> 
> Yeah I think so if you don't mind

did it, 
removed the unnecessary aditional  command if user skips submodule, as the 'make prepare' will fix it.

PR #536, opened 2024-09-10
Fix: tests

RaulTrombin:
The current 'make test' encounters some errors during execution.

The first commit addresses only the clippy [warnings](https://rust-lang.github.io/rust-clippy/master/index.html#/derivable_impls).

The second commit, [fix: Fix test, extract response on client error](https://github.com/paperclip-rs/paperclip/commit/6c141b5b46905d05bc1626e7afe4340af00c7f35), resolves the following error:

```
error[E0599]: no method named `into_inner` found for associated type `<C as ApiClient>::Response` in the current scope
  --> ./cli.rs:74:49
   |
74 |         Err(ApiError::Failure(_, _, r)) => Ok(r.into_inner()),
   |                                                 ^^^^^^^^^^ method not found in `<C as ApiClient>::Response`

```

Could someone confirm if this commit is the aproach to resolve the issue?

### DIFF
--- a/core/src/v2/models.rs
+++ b/core/src/v2/models.rs
@@ -31,9 +31,10 @@ static PATH_TEMPLATE_REGEX: Lazy<Regex> =
 const SPECIAL_HEADERS: &[&str] = &["content-type", "accept", "authorization"];
 
 /// OpenAPI version.
-#[derive(Clone, Copy, Debug, Serialize, Deserialize, Eq, PartialEq)]
+#[derive(Clone, Copy, Debug, Default, Serialize, Deserialize, Eq, PartialEq)]
 pub enum Version {
     #[serde(rename = "2.0")]
+    #[default]
     V2,
 }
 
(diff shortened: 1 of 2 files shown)


core/src/v2/models.rs
```diff
@@ -31,9 +31,10 @@ static PATH_TEMPLATE_REGEX: Lazy<Regex> =
 const SPECIAL_HEADERS: &[&str] = &["content-type", "accept", "authorization"];
 
 /// OpenAPI version.
-#[derive(Clone, Copy, Debug, Serialize, Deserialize, Eq, PartialEq)]
+#[derive(Clone, Copy, Debug, Default, Serialize, Deserialize, Eq, PartialEq)]
```
tiagolobocastro, line 34: Let's keep default as is for now to avoid bumping msrv?

tiagolobocastro: Or maybe it's a good time to bump it, though we'd need to have another breaking release but that's ok I guess

PR #541, opened 2024-09-18
Use re-export of serde_json in macros

### LINKED ISSUE #540: Use re-export of serde_json in macros
The `v2` macros generate code that directly uses the crate `serde_json`. The problem is that the user code might not have this crate as a dependency.

So, they might get confusing errors (about a missing `serde_json` crate when there is no instance of `serde_json` string in their project), and in order for the generated code to compile, they have to add it as a dependency which may be picked up as "unused" by linting tools.

From what I gathered writing some proc-macro crates, the good practice is to depend yourself on the crate, re-export the symbols you need and reference it from your own namespace.

Sufflope:
Fixes #540

### DIFF
--- a/core/src/v2/mod.rs
+++ b/core/src/v2/mod.rs
@@ -22,6 +22,7 @@ pub use self::{
     schema::Schema,
 };
 pub use paperclip_macros::*;
+pub use serde_json;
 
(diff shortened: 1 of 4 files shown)


tiagolobocastro (approved):
Thanks @Sufflope!

PR #546, opened 2024-10-05
fix: bump paperclip-macros minor

### LINKED ISSUE #542: `paperclip-macros` patch version `0.6.4` introduced a breaking change with min/max
This patch version changes the `DefaultSchemaRaw` derived struct by adding properties to it in commit 135096f9fd0c72a42929b4e188d4894d6b36fd56.

My problem is that I have a proc-macro crate where I generate code, including a hand-written `Apiv2Schema` `impl` for the type:

```rust
impl paperclip::v2::schema::Apiv2Schema for #ident {
    fn name() -> Option<String> {
        Some(stringify!(#ident).to_string())
    }
    fn raw_schema() -> paperclip::v2::models::DefaultSchemaRaw {
        paperclip::v2::models::DefaultSchemaRaw {
            name: Self::name(),
            reference: None,
            title: None,
            description: Some(#description.to_string()),
            data_type: Some(paperclip::v2::models::DataType::String),
            format: None,
            example: Some(serde_json::json!(#example)),
            properties: ::std::collections::BTreeMap::new(),
            items: None,
            enum_: vec![],
            extra_props: None,
            required: ::std::collections::BTreeSet::new(),
        }
    }
}
```

The issue is that I (and I guess other `paperclip` users) only add a dependency to the base crate, not the macros directly, The macro crate is pulled transitively, and `0.6.4` is semver-compatible with `0.6.3`, so if I run `cargo update` in my end project (which depends on `paperclip` and my internal crate) to update my dependencies for housekeeping, `paperclip-macros` gets updated to `0.6.4` but then my project does not compile anymore because my proc-macro crate generates incomplete `DefaultSchemaRaw` structs.

IDK what the course of action would be? I would say yank `paperclip-macros` `0.6.x` for `x > 3` and bump it to `0.7.0`, and release patch versions of `paperclip` for those which need those min/max params, with a dependency to `paperclip-macros` `0.7`?

tiagolobocastro:
This is to address min/max being added and breaking back compat. paperclip-macros 0.6.4 and 0.6.5 will be yanked from crates. paperclip 0.9.0 and 0.9.1 will also be yanked from crates.

Resolves #542

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.9.1] - 2024-09-10
+## [0.9.2] - 2024-10-13
+### Fixed
+- Switch to pro-macro-error2. [PR#545](https://github.com/paperclip-rs/paperclip/pull/545)
+- Bumps paperclip-macros to 0.7.0 to fix breaking change in 0.6.4. [PR#546](https://github.com/paperclip-rs/paperclip/pull/546)
+
+## [0.9.1] - 2024-09-10 [ YANKED ]
 ### Fixed
 - Support array field type. [PR#531](https://github.com/paperclip-rs/paperclip/pull/531)
 
-## [0.9.0] - 2024-09-07
+## [0.9.0] - 2024-09-07 [ YANKED ]
 ### Fixed
 - Support latest openapiv3 and indexmap. [PR#507](https://github.com/paperclip-rs/paperclip/pull/507)
 - Clippy useless vec lint. [PR#511](https://github.com/paperclip-rs/paperclip/pull/511)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "paperclip"
-version = "0.9.1"
+version = "0.9.2"
 edition = "2018"
 description = "OpenAPI tooling library for type-safe compile-time checked HTTP APIs"
 documentation = "https://paperclip-rs.github.io/paperclip/paperclip"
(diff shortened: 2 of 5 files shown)


tiagolobocastro:
@Sufflope
I think we also yank paperclip 0.9.0 and 0.9.1 unfortunately, because those need the min/max change.
So once the paperclip-macros 0.6.{4,5} is yanked then paperclip 0.9.{0,1} become broken anyway?
Is this the right thing to do?
CC 
@formlogic-kirk @snaggen  @RaulTrombin @thomedw 
 for visibility since you've raised PRs recently, see #542 for the problem we're trying to address here, sorry for the mess-up :(

tiagolobocastro:
I wonder if another way is to release 0.6.6 without min/max
Then pre 0.9.0 would get the correct code when updating, and existing 0.9.{0,1} would stay working.

Sufflope:
First let me say that I don't have previous experience with yanking, sorry (it's not a brag "I don't break things" :sweat_smile:  on the contrary, it's just that I never got to maintain a public library :cry: ), so I don't have prior battle-tested knowledge.

So my not-so-educated opinion is: the problem with your second approach is in the (contrived, admittedly, but technically possible) case where someone would have pinned a `=0.9.0` paperclip dependency but no explicit macros version, it would break on `cargo update`? So I think the safest option is your first one.

tiagolobocastro:
Won't it break anyway with the first approach? Because we'd be yanking the paperclip-macros pointed to by the paperclip 0.9.{0,1}?

Sufflope:
According to https://doc.rust-lang.org/cargo/commands/cargo-yank.html if I understand correctly, it should prevent new projects from adding paperclip 0.9.{0,1} as dependencies, but projects which have them already should continue compiling (with a warning telling them those deps are yanked). But at least users would be warned, and be able to update by switching to at least 0.9.2, instead of updating only the macro and getting a cryptic error that some fields are missing in a proc-macro. At least that's my understanding.

Sufflope:
Note that anyway on my side I updated all my internal projects to 0.9 and I generate min/max attributes, and I am quite aware of the problem we're discussing, so even if there's a subsequent problem with the chosen strategy, it won't be a big issue to me, so feel free to choose the strategy you think will work the best!

formlogic-kirk:
Looks good to me 👍

tiagolobocastro:
Alright cool, I'll try and do this tomorrow or over the weekend, thanks both

